PR #40, opened 2021-05-21
change TimeZoneOffsetFromUTC field of ios.AllValuesType from int to float64

SolskGaer:
according to the raw plist, the type of TimeZoneOffsetFromUTC is real, and it should be umarshaled to float64, when declared  as int, the retrieved value will always be zero.


danielpaulus (approved):
Thank you for the fix!

PR #41, opened 2021-06-01
Change forward to listen on all interfaces

YasharAzari:
Current forward configuration listens only to request from localhost.
This change allows us to access to the forwarded port from other origins.


danielpaulus:
> Current forward configuration listens only to request from localhost.
> This change allows us to access to the forwarded port from other origins.

Thanks :-)

PR #45, opened 2021-07-25
fix: META-INF file exists bug

wangchaoHZ:
**Bug Description:**
When i try to install a ipa file, i find if META-INF folder is included in the unzip IPA package. The following error messages will be generated.

"**{"err":"mkdir /var/folders/3_/fjr0f3dn58j22061qts7sssh0000gp/T/prefix693025361/META-INF: file exists","level":"fatal","msg":"failed writing","time":"2021-07-25T20:16:11+08:00"}**“

Therefore, i added code determine whether the folder exists before creating it.

### DIFF
--- a/ios/utils.go
+++ b/ios/utils.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"os"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


danielpaulus:
Nice catch! 
I should have though about that myself 🤦 

Thank you so much for contributing to go-ios :-D

PR #63, opened 2021-09-21
initial attempt to use com.apple.mobile.MCInstall to fetch profiles

testingbot:
Hi,

I'm trying to implement `com.apple.mobile.MCInstall` to fetch profiles from the device.
The goal is to list, remove and add profiles to the device, similar to the functionality in https://github.com/YueChen-C/py-ios-device/blob/main/ios_device/servers/mc_install.py

The problem is that I am not receiving anything from the plist codec decoder.

If someone could point me in the right direction, I can improve and expand this PR to make it work.


danielpaulus:
> Hi,
> 
> I'm trying to implement `com.apple.mobile.MCInstall` to fetch profiles from the device. The goal is to list, remove and add profiles to the device, similar to the functionality in https://github.com/YueChen-C/py-ios-device/blob/main/ios_device/servers/mc_install.py
> 
> The problem is that I am not receiving anything from the plist codec decoder.
> 
> If someone could point me in the right direction, I can improve and expand this PR to make it work.

Hey sorry for the late reply, for some reason I don't get notification Emails on PRs, I need to check my notification settings.
I'll take a look today :-)

danielpaulus:
@testingbot  took a little longer but I think essentially here is your problem:
```
func (mcInstallConn *Connection) HandleList() error {
	reader := mcInstallConn.deviceConn.Reader()
	request := map[string]interface{}{"RequestType": "GetProfileList"}
	bytes, err := mcInstallConn.plistCodec.Encode(request)
	if err != nil {
		return err
	}
	err = mcInstallConn.deviceConn.Send(bytes)
	if err != nil {
		return err
	}
	return mcInstallConn.readExchangeResponse(reader)
}
```
Usually you send dictionaries not Arrays. 
[]interface{}{} is basically a string array
`[]interface{}{"RequestType", "GetProfileList"}`

danielpaulus (approved):
I'll finish it later myself

PR #75, opened 2021-11-23
change permission for usbmuxd proxy socket

aluedeke:
change the permissions for the proxy daemon so all users can access it


danielpaulus (approved):
nice one!

PR #98, opened 2022-01-09
Add ability to make IORegEntry request

alyyousuf7:
Signed-off-by: Ali Yousuf <aly.yousuf7@gmail.com>

This does 3 things:
- Move `MobileGestalt` to its own file to separate the concern, including its request payload.
- It removes sending `goodBye()` message request, otherwise you cannot reuse the Diagnostics connection to make another request. User can use `defer diagnosticsConn.Close()` to close the connection it manually.
- Adds query function for `IORegEntry` request from diagnostics, which returns very useful data, such as Battery information with `AppleSmartBattery` as `key`.

### DIFF
--- a/ios/diagnostics/diagnostics.go
+++ b/ios/diagnostics/diagnostics.go
@@ -2,6 +2,7 @@ package diagnostics
 
 import (
 	"fmt"
+
 	ios "github.com/danielpaulus/go-ios/ios"
 )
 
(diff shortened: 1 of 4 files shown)


danielpaulus:
Really cool change, thanks :-) 
If you want, open a discussion or connect on LinkedIn. I'd love to learn about what you are doing with go-iOS and if I can help.

compoundradius:
How do you perform this query? Obtaining battery stats similar to
```ideviceinfo -u <udid> -q com.apple.mobile.battery'' is exactly what I am looking to replicate.

PR #129, opened 2022-04-21
Fix automatic image mounting on Windows

shamanec:
* This is meant to fix automatic image mounting on Windows OS  
* Added a check for the current OS to provide proper imageToFind string


danielpaulus (approved):
thanks!
I think there is also a bug with the latest iOS. Will look into testing this better soon,.

PR #140, opened 2022-06-14
When one forwarding tunnel connection closes, close the other as well.

briankrznarich:
Issue: port-forwarding clients hang indefinitely if/when the target device port closes.

At the end of io.Copy(deviceConn,clientConn), only one of the two connections is
guaranteed to be closed. If deviceConn fails (because an application on the
iOS device was shut down, crashed, or closed a connection), the client connection is
left open.  The effect is that the client will hang indefinitely, continuing to send
traffic we cannot forward, and listening for traffic that will never arrive.


danielpaulus (approved):
Thanks!

PR #144, opened 2022-06-30
fix some bug when install and run xcuitest & supoort context in forward and xcuitestrunner & add fsync

wangchaoHZ:
hi, daniel:
from this pull request, the main changes are as follows:

1、fix install ipa failed below iOS 10 devices
2、fix some bug when run xcuitest
3、support context in forward，we can cancel it，and export the StartNewProxyConnection function for use
4、support context in xcuitestrunner，so that we can cancel it if need
5、fix: house_arrest connection close failed when it is nil（If we run the enterprise signed XCUITEST ipa, this problem will occur）
6、The most important thing is support fsync command line, we can transfer or operate file with the iPhone，it support push、pull、remove、treeview、mkdir and so on.（Thanks for @[codeskyblue](https://github.com/alibaba/taobao-iphone-device/commits?author=codeskyblue) contribution in tidevice）

such as we can use 'fsync push' to push file to iPhone device，like this:
```
./go-ios fsync push --srcPath="src path" --dstPath="dst path"
```
use 'fsync pull' to pull file from iPhone device, like this:
```
./go-ios fsync pull --srcPath="src path" --dstPath="dst path"
```
or do 'fsync remove'  'fsync tree'  'fsync mkdir' in target path or file.


danielpaulus:
Wow this is really great work! I am checking it right now!

ios/dtx_codec/connection.go
```diff
@@ -18,7 +18,7 @@ type MethodWithResponse func(msg Message) (interface{}, error)
 type Connection struct {
 	deviceConnection       ios.DeviceConnectionInterface
 	channelCodeCounter     int
-	activeChannels         map[int]*Channel
+	activeChannels         sync.Map
```
danielpaulus, line 21: 🤦  yeah sync.Map.. why didn't I think of that.

PR #147, opened 2022-07-23
Add ability to enable/disable AssistiveTouch soft "home" button. iOS 11+

briankrznarich:
Using the same technique already used for querying and setting the device language and locale, go-ios can now enable and disable AssistiveTouch using the the magic boolean com.apple.Accessibility.AssistiveTouchEnabledByiTunes

This commit slightly modifies the existing SetValueForDomain() and related methods and structs to take an interface{} value (rather than a string) in order to pass a raw bool to lockdown.

This commit also replaces a few inconsistently used tabs in the verbose command-line "help" output.

Thank you to the team at alibaba/taobao-iphone-device (tidevice) for sharing the magic word to make this happen.

Validated on iPhone 5s, 6, 6s, 6s plus, 7, 8, se 1st gen, se 2nd gen, and 11 pro max.  iOS 11.4.1, 12.1, 12.5.5, 13.4, 13.6, 14.0.1, 14.7, 15.0, 15.3, and 16.0.1 beta.

Failed on an iPhone 6+ running iOS 10.2.1 (Programmed to exit with custom error including hint: “Is this device running iOS 11+?”)

### DIFF
--- a/ios/lockdown_assistivetouch.go
+++ b/ios/lockdown_assistivetouch.go
@@ -1,0 +1,52 @@
+package ios
+
+import log "github.com/sirupsen/logrus"
+import "fmt"
+
+const accessibilityDomain = "com.apple.Accessibility"
+const assistiveTouchKey  = "AssistiveTouchEnabledByiTunes"
+
+// EnableAssistiveTouch creates a new lockdown session for the device and enables or disables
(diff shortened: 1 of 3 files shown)


briankrznarich:
Going to go fix those tests...

PR #148, opened 2022-07-23
Add ability to enable/disable AssistiveTouch soft "home" button. iOS 11+

briankrznarich:
Using the same technique already used for querying and setting the device language and locale, go-ios can now enable and disable AssistiveTouch using the the magic boolean com.apple.Accessibility.AssistiveTouchEnabledByiTunes

This commit slightly modifies the existing SetValueForDomain() and related methods and structs to take an interface{} value (rather than a string) in order to pass a raw bool to lockdown.

This commit also replaces a few inconsistently used tabs in the verbose command-line "help" output.

Thank you to the team at alibaba/taobao-iphone-device (tidevice) for sharing the magic word to make this happen.

Validated on iPhone 5s, 6, 6s, 6s plus, 7, 8, se 1st gen, se 2nd gen, and 11 pro max.  iOS 11.4.1, 12.1, 12.5.5, 13.4, 13.6, 14.0.1, 14.7, 15.0, 15.3, and 16.0.1 beta.

Failed on an iPhone 6+ running iOS 10.2.1 (Exits with error including hint: “Is this device running iOS 11+?”)

### DIFF
--- a/ios/lockdown_assistivetouch.go
+++ b/ios/lockdown_assistivetouch.go
@@ -33,7 +33,8 @@ func GetAssistiveTouch(device DeviceEntry) (bool, error) {
 
 	if err != nil {
 		return false, err
-	}else if enabledIntf == nil{
+	}
+	if enabledIntf == nil{
 		// In testing, nil was returned in only one case, on an iOS 14.7 device that should already have been paired.
 		// Calling SetAssistiveTouch() directly returned the somewhat more useful error: SetProhibited
 		// After re-running "go-ios pair", full functionality returned.
@@ -42,7 +43,7 @@ func GetAssistiveTouch(device DeviceEntry) (bool, error) {
 	enabledUint64, ok := enabledIntf.(uint64)
 	if !ok {
 		// On iOS 10.x at least, "false" is returned, perhaps for any key at all.  Attempting to manipulate AssistiveTouchEnabledByiTunes had no effect
-		return false, fmt.Errorf("Expected unit64 0 or 1 when querying %s.%s, but received %T:%+v. Is this device running iOS 11+.", accessibilityDomain, assistiveTouchKey, enabledIntf, enabledIntf)
+		return false, fmt.Errorf("Expected unit64 0 or 1 when querying %s.%s, but received %T:%+v. Is this device running iOS 11+?", accessibilityDomain, assistiveTouchKey, enabledIntf, enabledIntf)
 	} else if enabledUint64 != 0 && enabledUint64 != 1{
 		// So far this has never happened
 		return false, fmt.Errorf("Expected a value of 0 or 1 for %s.%s, received %d instead!", accessibilityDomain, assistiveTouchKey, enabledUint64)
--- a/ios/lockdown_assistivetouch.go
+++ b/ios/lockdown_assistivetouch.go
@@ -1,0 +1,52 @@
+package ios
+
+import log "github.com/sirupsen/logrus"
+import "fmt"
+
+const accessibilityDomain = "com.apple.Accessibility"
+const assistiveTouchKey  = "AssistiveTouchEnabledByiTunes"
+
+// EnableAssistiveTouch creates a new lockdown session for the device and enables or disables
+// AssistiveTouch (the on-screen software home button), by using the special key AssistiveTouchEnabledByiTunes.
+// Setting to true will enable AssistiveTouch
+// Setting to false will disable AssistiveTouch, regardless of whether it was previously-enabled through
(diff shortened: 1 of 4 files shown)


ios/lockdown_assistivetouch.go
```diff
@@ -0,0 +1,52 @@
...
+	}else if enabledIntf == nil{
+		// In testing, nil was returned in only one case, on an iOS 14.7 device that should already have been paired.
+		// Calling SetAssistiveTouch() directly returned the somewhat more useful error: SetProhibited
+		// After re-running "go-ios pair", full functionality returned.
+		return false, fmt.Errorf("Received null response when querying %s.%s. Try re-pairing the device.", accessibilityDomain, assistiveTouchKey)
+	}
+	enabledUint64, ok := enabledIntf.(uint64)
+	if !ok {
+		// On iOS 10.x at least, "false" is returned, perhaps for any key at all.  Attempting to manipulate AssistiveTouchEnabledByiTunes had no effect
+		return false, fmt.Errorf("Expected unit64 0 or 1 when querying %s.%s, but received %T:%+v. Is this device running iOS 11+.", accessibilityDomain, assistiveTouchKey, enabledIntf, enabledIntf)
```
danielpaulus, line 45: I think it would be cool if you could add a check in the beginning for the iOS version. 
If we already know it doesn't work on iOS10 maybe better to fail fast?
In either case, including the device version in the error message could be good. 

Here is an example how I work with ios versions: https://github.com/danielpaulus/go-ios/blob/0e140a0f03560ce1302f00e70348796cd525dfd0/ios/imagemounter/imagedownloader.go#L56

wdyt?

ios/lockdown_assistivetouch.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer lockDownConn.Close()
+
+	enabledIntf, err := lockDownConn.GetValueForDomain(assistiveTouchKey, accessibilityDomain)
+
+	if err != nil {
+		return false, err
+	}else if enabledIntf == nil{
```
danielpaulus, line 36: please don't use else if. 
```
if ... {
return ..
}
if enabledIntf==nil{
..
}
```
Does the same with less nesting. Usually it is easier to read.

main.go
```diff
@@ -857,6 +881,33 @@ func language(device ios.DeviceEntry, locale string, language string) {
 	fmt.Println(convertToJSONString(lang))
 }
 
+func assistiveTouch(device ios.DeviceEntry, operation string) {
+	wasEnabled, err := ios.GetAssistiveTouch(device)
+	var enable bool
+	exitIfError("failed getting current AssistiveTouch status", err)
+
+	if operation == "enable" {
```
danielpaulus, line 889: this might be nicer as a switch case like f.ex.
```
func main() {
    day := "Tue"
  
    // Use switch on the day variable.
    switch {
    case day == "Mon":
        fmt.Println("Monday")
    case day == "Tue":
        fmt.Println("Tuesday")
    case day == "Wed":
        fmt.Println("Wednesday")
    }
}
```

danielpaulus (changes requested):
Really nice work! I only have a few nitpicks.

ios/lockdown_assistivetouch.go
briankrznarich: I've gone ahead and added a fail-fast check, with a more to-the-point error message, on the command-line invocation of go-ios.    However, I also added a "--force" override to go ahead and try the operations anyway, since I've only tested on a single device older than iOS 11 (and have not tested on iPad).  At least for a little while, it would be nice to *try* this method on older hardware.

I left the "Is this device running iOS 11+?" in the library call error message, with the idea that this provides some feedback for non-command-line invocations without making it impossible to invoke the function if you want to try.  
Presumably if you are using this in a library with a mechanism to automate assistive touch, you should have written in custom case-checking for older devices yourself anyway (presuming you intend to support iOS 10 and older...). 

What do you think?

danielpaulus (changes requested):
Great stuff, I have some minor nitpicking still. In general I try to only use 
else or else if constructs if absolutely necessary. I think they make the code more difficult to read and prefer "guard style" if clauses that terminate the function with a return statement. 
Thanks for enduring my nitpicking :-D

ios/lockdown_assistivetouch.go
danielpaulus: Perfect.

danielpaulus:
I'll merge it for the sake of getting it in. The nitpicks I can address later or sth. :-)
Thanks a lot for this contribution!

PR #150, opened 2022-07-24
Add option to pretty-print JSON command output. Add trailing newlines…

briankrznarich:
This commit adds a the --pretty option to format any JSON output with a 4-space indent using golang's json.MarshalIndent, making certain output easier to read when working manually on the command-line.

While updating a few printf calls, I also noticed a few commands that I believe were unintentionally omitting a trailing newline, so that after running those commands the user's command prompt would be at the end of the output of the previous command.  In principle this *could* break scripted commands that assume no newline? Hard to imagine...

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,7 @@ import (
 
 //JSONdisabled enables or disables output in JSON format
 var JSONdisabled = false
+var prettyJSON = false
 
 func main() {
 	Main()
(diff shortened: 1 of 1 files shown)


briankrznarich:
There is one other alteration in this PR, which maybe I should have been more careful about.
<  --nojson       Disable JSON output (default).
>  --nojson       Disable JSON output

In my experience, when "(default)" appears next to an option, that usually indicates the option is active by default.  Thus, I naturally interpret "--nojson Disable JSON output(default)" as "JSON disabled by default", which of course it isn't.  

But, then, why have that option at all...  I now see that maybe the intention was something more like "Disables the default JSON output".

In any case, I thought the behavior of the application must have changed some time in the past, without updating the help-documentation.  Thus, while working on other JSON settings I snipped out the "(default)" just to clean up the docs.  If this was not a welcome change, my apologies.

danielpaulus (approved):
Good stuff! 
I think the default was put there by accident. I wrote a few pieces of go-ios in the car, while constantly being distracted, so there are some quirks still. Thanks for removing this!

PR #152, opened 2022-07-24
Add command-line option to kill processes by process-name or PID

briankrznarich:
This commit also adds a "--system" option that allows built-in processes, such as com.apple.Maps, to be killed by bundleID.  (When specifying --pid or --process, there are no restrictions)

Examples of processes that one might want to kill this way:
com.apple.Maps   (using the --system option)
replayd          (when broadcast streaming becomes irrevocably hung)
assistivetouchd  (useful as a test.  If the AssistiveTouch home button is active, killing this makes the button briefly disappear, but it restarts automatically).

This commit attempts to preserve all previous behavior, but a few changes have been made:
* Some typos were corrected ("killd" -> "killed")
* Some spaces were added in log messages where variables were concatenated with no spaces
* The "kill" command now exits with a non-zero exit status for any failure to kill a process. previously log.Error() was called, followed by exit(0).  Now it should be possible to tell whether a process was actually killed based on the exit status of the application.

Note: Currently, using "--system" *prevents* killing user applications.  After PR #151, replacing a call to "BrowseSystemApps()" with the new "BrowseAllApps()" will enable killing by bundle ID without considering how the app was installed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,7 +92,7 @@ Usage:
   ios uninstall <bundleID> [options]
   ios apps [--system] [--all] [options]
   ios launch <bundleID> [options]
-  ios kill (<bundleID> | --pid=<processID> | --process=<processName>) [--system] [options]
+  ios kill (<bundleID> | --pid=<processID> | --process=<processName>) [options]
   ios runtest <bundleID> [options]
   ios runwda [--bundleid=<bundleid>] [--testrunnerbundleid=<testbundleid>] [--xctestconfig=<xctestconfig>] [--arg=<a>]... [--env=<e>]... [options]
   ios ax [options]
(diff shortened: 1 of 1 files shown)


briankrznarich:
Right now (and prior to this PR), if you call:
go-ios <bundleid>
the bundle id must be in the list returned by BrowseUserApps(). ie. it only works on user-installed applications.

I wonder, is there a reason for this behavior? "kill" is capable of killing other bundle ids, and surely if the user tries to kill, say, "com.apple.Maps", they know that's what they want to do.

In PR#151 I added a function BrowseAllApps().  If using that function instead of BrowseUserApps(), the user could attempt to kill anything with an identifiable bundle id.

I've generally tried to maintain backward-compatible behavior, but what do you think of expanding "kill <bundleID>" to simply attempt to kill anything, without the use of additional flags? (This PR adds a "--system" option to kill, I'd be happy to adjust the PR if #151 is accepted and remove the "--system" option entirely).

danielpaulus:
> I've generally tried to maintain backward-compatible behavior, but what do you think of expanding "kill " to simply attempt > to kill anything, without the use of additional flags? (This PR adds a "--system" option to kill, I'd be happy to adjust the PR > if https://github.com/danielpaulus/go-ios/pull/151 is accepted and remove the "--system" option entirely).

I agree. This is a really good idea. Please go ahead and remove the system option so kill just kills whatever PID you give it.

danielpaulus (approved):
thanks! This is a really good improvement.

PR #164, opened 2022-08-19
Exit with error when attempting to forward an unavailable port.

briankrznarich:
Previous code has an out-of-order error check, starting a thread and *then* returning on error.  This leads to a null-pointer panic if you attempt to forward a port that is already in use.


danielpaulus (approved):
Thanks!

PR #166, opened 2022-08-24
Add/basic api

danielpaulus:
Adds the boilerplate and first endpoints for the new REST-API using [gin gonic](https://github.com/gin-gonic/gin) as a framework. 
- middleware for streaming responses, like syslog and listen
- middleware to limit concurrent access on a udid basis 
- middleware to parse udid from path
- swagger setup to write open api docs
First endpoints:
- list installed apps
- list devices
- screenshot
- device info
- syslog
- listen


restapi/main.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"github.com/danielpaulus/go-ios/restapi/api"
+	_ "github.com/danielpaulus/go-ios/restapi/docs"
```
shamanec, line 5: This folder is currently on .gitignore so build fails. Easily fixed by removing the import since its not used but shouldn't we push the `docs` folder to the repo?

restapi/api/server.go
```diff
@@ -0,0 +1,88 @@
...
+
+	err := router.Run(":8080")
+	if err != nil {
+		log.Error(err)
+	}
+}
+
+func HeadersMiddleware() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		c.Writer.Header().Set("Content-Type", "text/event-stream")
```
shamanec, line 82: Probably most endpoints will not be SSE, maybe we should add this header only where needed and not in the generic middleware?

restapi/api/server.go
```diff
@@ -0,0 +1,88 @@
...
+		c.Writer.Header().Set("Cache-Control", "no-cache")
+		c.Writer.Header().Set("Connection", "keep-alive")
+		c.Writer.Header().Set("Transfer-Encoding", "chunked")
```
shamanec, line 85: Probably most endpoints will not be SSE, maybe we should add these headers only where needed and not in the generic middleware?

restapi/api/general_device_endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+func Screenshot(c *gin.Context) {
+	udid := c.Param("udid")
+	device, _ := ios.GetDevice(udid)
+
+	conn, err := screenshotr.New(device)
+	log.Error(err)
+	b, _ := conn.TakeScreenshot()
+
+	c.Header("Content-Type", "image/png")
+	c.Data(http.StatusOK, "application/octet-stream", b)
```
shamanec, line 68: Is there a specific reason this is not `image/png` but `application/octet-stream`?

shamanec: Don't mind me, it seems I don't understand how this works, response content-type header is `image/png`

restapi/api/server.go
```diff
@@ -0,0 +1,88 @@
...
+	device := v1.Group("/device/:udid")
+	device.GET("/info", Info)
+	device.GET("/screenshot", Screenshot)
+
+	v1.GET("/listen", Listen)
+	v1.GET("/list", List)
+
+	router.GET("/swagger/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
+
+	err := router.Run(":8080")
```
shamanec, line 74: We should provide an option for user selected port, maybe using `flag`, `8080` is quite generic and could already be used by something else

danielpaulus:
@shamanec  Thanks for the review! Before reading&responding to your questions, I wanted to say this is me figuring the boilerplate still. This whole thing is deployed to a local linux laptop and the tests run on Checkly. 
I'll respond now, but we probably should meet and discuss this work in the coming week.

restapi/main.go
danielpaulus: It is auto generated by go-swagger. At some point a Makefile or sth similar should probably be provided. 
Take a look here: https://github.com/danielpaulus/go-ios/blob/add/basicAPI/.github/workflows/deploy.yml
at the preliminary build config.
TL;DR
run
`go install github.com/swaggo/swag/cmd/swag@latest
          && swag init`
and it will work.

restapi/api/server.go
danielpaulus: agreed.

restapi/api/server.go
danielpaulus: agreed.
I have not yet thought about how to manage configuration. But I think also other things will need to be configured later on. One good example is rate limiting. Some endpoints will have to be rate limited so they cannot be executed concurrently, or only with a limit. F.ex. hitting the screenshot endpoint 20 times in parallel will probably kill it, or we cache it somehow.

restapi/main.go
shamanec: My bad, I didn't check the workflow, but I already had swaggo so just generated them locally for myself. Good stuff!

restapi/api/general_device_endpoints.go
danielpaulus: It's fine, I am also a complete noob with gin-gonic :-D 
It is very much possible I am doing this wrong.

Issue #169, opened 2022-08-29
Developer Disk Image is not saved in the proper path


k-inoway:
**Describe the bug**
Developer Disk Image is not saved in the proper path. The OS version folder is outside the devimages folder.

See screenshot for details.

**To Reproduce**
1. run `ios image auto`

**Expected behavior**
`devimages\15.6.1\`

**Screenshots**
![image](https://user-images.githubusercontent.com/40589931/187114829-5e015e38-64b9-48a6-950e-d76c796e5429.png)

**go-ios version**
 - v1.0.80

**Desktop (please complete the following information):**
 - OS: Windows10 20H2

**Smartphone (please complete the following information):**
 - Device: iPhone12 mini
 - OS: iOS 15.6.1

PR #170, opened 2022-08-29
Fix image auto download path without basedir

shamanec:
This is a fix for #169. After the recent changes of the images repo, automatic mounting without `basedir` flag was creating `devimages` folder but was saving the images in the base dir of the app. Changed the way `devimages` is referenced according to the latest changes so the images are saved inside it when no `basedir` flag is provided.

Tested on Windows and Ubuntu 18.04 with and without `basedir`


danielpaulus (approved):
Thanks!

Issue #182, opened 2022-10-08
Add lock endpoints to REST API 


danielpaulus:
Go-iOS just got a new REST API. It needs a mechanism to ensure certain endpoints can be reserved with a lock to prevent multiple users using the same device at the same time. 

https://github.com/danielpaulus/go-ios/tree/main/restapi/README.md for how to get started.

I think the solution should come in 2 PRs:
First PR:
1. create lock_endpoints.go
2. add GET /locks which should return all current locks like this:
 ```
[
{udid:"udidofdevice", lock:"9C9DEB66-0EF1-432F-BA5B-038066C21DCD", "lastUsed":1665227406347},
....
]
3. add POST /lock/{udid} endpoint which will generate and store a randomly generated UUID in a sync.Map and return it or fail if the device already has a lock
4. add DELETE /lock/{udid} that allows to remove a lock
5. add integration test for the API
```
Second PR:
1. create middleware that will check if the request header contains a X-GO-IOS-LOCK field with a UUID
2. check if the sync.Map contains the lock UUID for the specified device UDID and return errors if it doesn't
3. update the lastUsed field to time.now()
4. add a const ADMIN_LOCK in the code that will always work for every device
5. create a test for the middleware

ash-kamrip:
@danielpaulus   I would like to work on this , if not all i think I can probably work on 1st or 2nd PR. Could you count these as my contributions to hacktoberfest ?  Also , right now I'm going through ios to see how to implement this

danielpaulus:
@ash-kamrip Cool! Yes, every one of the 3 PRs will be good for Hacktoberfest :-) Depending on how much code it will be, you can also split them into smaller ones. My 3 PRs were more of a suggestion. 
You only have to check out the code in here https://github.com/danielpaulus/go-ios/tree/main/restapi
You can ignore the rest of the codebase for this issue.

ash-kamrip:
@danielpaulus  @shamanec I think the work here is done , I am new to creating api's , I was thinking about how to protect existing endpoints/functions by wrapping `lock` and `unlock` around the code base but @shamanec 's approach is quite new for me , thanks for the learning 😄

danielpaulus:
@ash-kamrip no worries. If you like, you can continue helping us out a bit with adding more endpoints. Of course Hacktoberfest is pretty much over, so I understand if not. 
I'll be adding more issues for the remaining endpoints in the coming days.

PR #187, opened 2022-10-22
Add lock endpoints to reserve devices

shamanec:
This is the first batch of code for #182  

* Added POST `/reserve/:udid` endpoint to reserve a device by provided UDID
* Added DELETE `/reserve/:udid` endpoint to release a reserved device by provided UDID
* Added GET `/reserved-devices` endpoint to get a list of the currently reserved devices
* Added a goroutine that will check reserved devices every minute and remove devices that were not used for more than 5 minutes
* Added API tests for the new package, not really sure if that's what was meant by 'integration tests'

I took the liberty to change the naming convention a bit since we reserve devices, not actually `lock` them. I used a normal map instead of a `sync.Map` because I think you can have limited number of devices on a physical host and based on the `go-ios` features you probably can't have that much of concurrency. Let me know what you think and I can try re-writing it


restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/gin-gonic/gin"
+)
+
+var reservedDevicesMap = make(map[string]*reservedDevice)
+var reserveMutex sync.Mutex
```
danielpaulus, line 14: I think with a sync.Map, you would not need to use the Mutex

restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Every minute loop through the map of reserved devices and check if a reserved device last used timestamp was more than 5 minutes(300000 ms) ago
+	// If any, remove them from the map
+	for range time.Tick(time.Second * 60) {
+		reserveMutex.Lock()
+		for udid, reservedDevice := range reservedDevicesMap {
+			currentTimestamp := time.Now().UnixMilli()
+			diff := currentTimestamp - reservedDevice.LastUsedTimestamp
+
+			if diff > 300000 {
```
danielpaulus, line 34: nit: is something like time.Minute*5 possible here?
also would be good to have it in a const, I am planning to make this configurable later

restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+// @Router       /reserve/:udid [post]
+func ReserveDevice(c *gin.Context) {
+	udid := c.Param("udid")
+	reservationID := randomReservationID()
+
+	reserveMutex.Lock()
+	defer reserveMutex.Unlock()
+
+	// Check if there is a reserved device for the respective UDID
+	device := reservedDevicesMap[udid]
```
danielpaulus, line 58: you can use device, exists := reservedDevicesMap[udid] instead of a nil check

restapi/api/reservation/reserve_endpoints.go
danielpaulus: But I have no strong opinion, the code looks good. No need to change it imo

restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+// @Produce      json
+// @Success      200  {object} []reservedDevice
+// @Router       /reserved-devices [get]
+func GetReservedDevices(c *gin.Context) {
+	reserveMutex.Lock()
+	defer reserveMutex.Unlock()
+
+	var reserved_devices []reservedDevice
+	if len(reservedDevicesMap) == 0 {
+		c.IndentedJSON(http.StatusOK, reservedDevice{Message: "No reserved devices found"})
```
danielpaulus, line 110: I think it would be better to return `[]` empty list here for consistency

restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+		return
+	} else {
```
danielpaulus, line 112: You don't really need the else here, right?
there is an 
if (..){
...
return
}
above. So else is not needed.

restapi/api/reservation/reserve_endpoints.go
```diff
@@ -0,0 +1,138 @@
...
+			})
+		}
+	}
+
+	c.IndentedJSON(http.StatusOK, reserved_devices)
+}
+
+const charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890"
+
+func randomReservationID() string {
```
danielpaulus, line 128: better to use UUIDs generated by a library like https://github.com/google/uuid
for this

danielpaulus (changes requested):
Nice work! I just have some minor complaints.

restapi/api/reservation/reserve_endpoints.go
shamanec: added an exported variable `ReservedDevicesTimeout` with default value of 5, can be accessed from main package to make it configurable

restapi/api/reservation/reserve_endpoints.go
shamanec: done

restapi/api/reservation/reserve_endpoints.go
shamanec: done, removed `else` statements in all places

restapi/api/reservation/reserve_endpoints.go
shamanec: good point, done

danielpaulus (approved):
Thanks alot for all the work!

PR #188, opened 2022-10-25
Add endpoints for changing device location

shamanec:
* Added `GenericResponse` struct to have a single interface that can be used to return both errors and meaningful messages where applicable, instead of generating JSON on each spot with `gin.H` or whatever. If you are okay with this I can later update the existing functions
* Added endpoint to change the current device location by `latitude` and `longtitude` query params  
* Added endpoint to reset the device location to the actual one  
* Bubbled up an unhandled error in my own code for device location simulation

### DIFF
--- a/ios/simlocation/simlocation.go
+++ b/ios/simlocation/simlocation.go
@@ -162,7 +162,10 @@ func SetLocationGPX(device ios.DeviceEntry, filePath string) error {
 				pointLat := point.PointLatitude
 
 				// Set the current point location by its latitude and longitude
-				SetLocation(device, pointLat, pointLon)
+				err = SetLocation(device, pointLat, pointLon)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
--- a/restapi/api/device_endpoints.go
+++ b/restapi/api/device_endpoints.go
@@ -95,9 +95,9 @@ func SetLocation(c *gin.Context) {
 	err := simlocation.SetLocation(device, latitude, longtitude)
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, GenericResponse{Error: err.Error()})
-	} else {
-		c.JSON(http.StatusOK, GenericResponse{Message: "Device location set to latitude=" + latitude + ", longtitude=" + longtitude})
+		return
 	}
+	c.JSON(http.StatusOK, GenericResponse{Message: "Device location set to latitude=" + latitude + ", longtitude=" + longtitude})
 }
 
 // Reset to the actual device location
@@ -113,7 +113,7 @@ func ResetLocation(c *gin.Context) {
 	err := simlocation.ResetLocation(device)
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, GenericResponse{Error: err.Error()})
-	} else {
-		c.JSON(http.StatusOK, GenericResponse{Message: "Device location reset"})
+		return
 	}
+	c.JSON(http.StatusOK, GenericResponse{Message: "Device location reset"})
 }
--- a/restapi/api/device_endpoints.go
+++ b/restapi/api/device_endpoints.go
@@ -1,12 +1,14 @@
 package api
 
 import (
+	"net/http"
+
 	"github.com/danielpaulus/go-ios/ios"
 	"github.com/danielpaulus/go-ios/ios/instruments"
 	"github.com/danielpaulus/go-ios/ios/screenshotr"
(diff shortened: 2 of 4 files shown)


ios/simlocation/simlocation.go
```diff
@@ -162,7 +162,10 @@ func SetLocationGPX(device ios.DeviceEntry, filePath string) error {
 				pointLat := point.PointLatitude
 
 				// Set the current point location by its latitude and longitude
-				SetLocation(device, pointLat, pointLon)
+				err = SetLocation(device, pointLat, pointLon)
```
danielpaulus, line 165: nice! I am also constantly looking for those. Go makes it too easy to miss those imo. 
And it can make the library/CLI tool really difficult to use.

restapi/api/device_endpoints.go
```diff
@@ -64,3 +66,54 @@ func Screenshot(c *gin.Context) {
...
+// @Param        latitude  query      string  true  "Location latitude"
+// @Param        longtitude  query      string  true  "Location longtitude"
+// @Success      200  {object}  GenericResponse
+// @Failure		 422  {object}  GenericResponse
+// @Failure		 500  {object}  GenericResponse
+// @Router       /device/{udid}/setlocation [post]
+func SetLocation(c *gin.Context) {
+	device := c.MustGet(IOS_KEY).(ios.DeviceEntry)
+	latitude := c.Query("latitude")
+	if latitude == "" {
```
danielpaulus, line 84: Very nice! I love good error messages!

restapi/api/device_endpoints.go
```diff
@@ -64,3 +66,54 @@ func Screenshot(c *gin.Context) {
...
+	longtitude := c.Query("longtitude")
+	if longtitude == "" {
+		c.JSON(http.StatusUnprocessableEntity, GenericResponse{Error: "longtitude query param is missing"})
+		return
+	}
+
+	err := simlocation.SetLocation(device, latitude, longtitude)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, GenericResponse{Error: err.Error()})
+	} else {
```
danielpaulus, line 98: I am not the biggest fan of else statements.
I usually prefer this pattern inspired by swift guards:
if err != nil{
...
return
}
rest of the code

restapi/api/routes.go
```diff
@@ -8,6 +8,8 @@ func registerRoutes(router *gin.RouterGroup) {
 	device.Use(DeviceMiddleware())
 	device.GET("/info", Info)
 	device.GET("/screenshot", Screenshot)
+	device.POST("/setlocation", SetLocation)
```
danielpaulus, line 11: set location is technically a `PUT` operation, right?

danielpaulus (changes requested):
This is great :-)

ios/simlocation/simlocation.go
shamanec: Yeah, I know... sometimes I just skip handling the error while testing out code and then completely forget about it.. 😆

restapi/api/routes.go
shamanec: Mmm I am not really sure `PUT` is applicable in this case because this is not a resource we are modifying, but either way is acceptable for me, let me know if you prefer to use `PUT`

restapi/api/device_endpoints.go
shamanec: I also usually do this, but my other development efforts get in the way :D Updated it in both endpoints

ios/simlocation/simlocation.go
danielpaulus: haha same for me 😅

restapi/api/device_endpoints.go
danielpaulus: thanks!

restapi/api/routes.go
danielpaulus: yes because PUT is an idempotent operation. Meaning, if I call it n-times with the same parameters, there will be no change in system state, which is the case here.
It does not matter if I call setlocation?lat=90&lon=90 once, twice or 500 times, the result will be the same.

danielpaulus: (idempotency is a really nice property to have in distributed systems, so the endpoint should be proud of it)

shamanec: that makes sense, done

danielpaulus (approved):
This is great, thanks for the contribution!

PR #190, opened 2022-10-27
Add endpoints for getting/enabling/disabling conditions

shamanec:
* Added endpoint to get the available conditions that can be enabled on a device  
* Added endpoint to enable a condition on a device by `profileTypeID` and `profileID` query params
* Added endpoint to disable the currently active condition on a device

### DIFF
--- a/restapi/api/device_endpoints.go
+++ b/restapi/api/device_endpoints.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"net/http"
+	"sync"
+	"time"
 
 	"github.com/danielpaulus/go-ios/ios"
 	"github.com/danielpaulus/go-ios/ios/instruments"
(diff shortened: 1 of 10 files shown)


restapi/api/device_endpoints.go
```diff
@@ -115,5 +117,147 @@ func ResetLocation(c *gin.Context) {
...
 	}
+
 	c.JSON(http.StatusOK, GenericResponse{Message: "Device location reset"})
 }
+
+//========================================
+// DEVICE STATE CONDITIONS
+//========================================
+
+var deviceConditionsMap = make(map[string]deviceCondition)
```
danielpaulus, line 128: I have added `LimitNumClientsUDID()` as a Middleware so you can limit the max. calls to a device specific endpoint to what makes sense. My idea was for every endpoint that should never be executed in parallel, to use that middleware.  Right now it is even hardcoded to 1 max. concurrent call. If you put that in front, you do not need a mutex and can simplify your code.

danielpaulus: In general for most functionality, allowing to call it in parallel is probably a bad idea as the behavior of the device would be undefined. Maybe the middleware must be even extended at some point to prevent calling f.ex. disable and enable at the same time.

danielpaulus (changes requested):

restapi/api/device_endpoints.go
shamanec: The purpose of this map is to hold the device conditions for all host devices that have them applied. In this case the Mutex is introduced to handle the concurrent access to the map for different devices enabling/disabling of conditions. I can still use that middleware but it will not simplify the current code :)

danielpaulus: So I think we need a middleware to limit concurrent access based on udid level. I cannot verify it, but I think just attaching the `LimitNumClientsUDID` middleware to a routergroup should already lock it down, so any endpoint in that group can only be called in series. 
Then you don't need a Mutex because there cannot be any concurrent calls. 
Doing this with such a Middleware is preferrable imo because it is less error prone than adding Mutexes to all endpoints.

restapi/api/routes.go
```diff
@@ -13,6 +13,9 @@ func registerRoutes(router *gin.RouterGroup) {
 	device.GET("/screenshot", Screenshot)
 	device.PUT("/setlocation", SetLocation)
 	device.POST("/resetlocation", ResetLocation)
+	device.GET("/conditions", GetSupportedConditions)
```
danielpaulus, line 16: I think if you put your three endpoints into an extra routergroup r
and then `r.Use(api.LimitNumClientsUDID())` you should not need a Mutex.

danielpaulus:
Merging it now, we can improve the concurrency later.

PR #193, opened 2022-11-02
Add reservation middleware

shamanec:
Added middleware for device reservation #182 
* Moved reservation code to the `api` package to make it easier to test and also there was no real point of separating it
* Added middleware that checks for `X-GO-IOS-RESERVE` header for endpoints on which it is applied
* Added a reservation admin UUID
* Reworked reservation endpoints to be more in line with the other device endpoints
* Added tests for the middleware and also cleaned up reservation tests code a bit


restapi/api/routes.go
```diff
@@ -17,22 +18,19 @@ func registerRoutes(router *gin.RouterGroup) {
 	device.PUT("/enable-condition", EnableDeviceCondition)
 	device.POST("/disable-condition", DisableDeviceCondition)
 
-	router.GET("/reserved-devices", reservation.GetReservedDevices)
-	reservations := router.Group("/reserve/:udid")
-	reservations.Use(DeviceMiddleware())
-	reservations.POST("/", reservation.ReserveDevice)
-	reservations.DELETE("/", reservation.ReleaseDevice)
+	device.POST("/reserve", ReserveDevice)
```
danielpaulus, line 21: let's call it `reservations` instead of `reserve` to be more RESTful. 
Then you can also rename /reserve-devices to just reservations, so it is more consistent.

restapi/api/routes.go
```diff
@@ -17,22 +18,19 @@ func registerRoutes(router *gin.RouterGroup) {
...
+	device.DELETE("/reserve", ReleaseDevice)
```
danielpaulus, line 22: I just realized this now, the DELETE endpoint needs to be connected to the UUID token you get for a reservation. Otherwise anyone could just delete my device reservations. 
so it should be DELETE /reservations/:token.
Technically you do not need the device UDID for it.

restapi/api/reserve_endpoints_test.go
```diff
@@ -122,9 +124,147 @@ func TestValidateDeviceNotReserved(t *testing.T) {
...
+	launchAppRequest, err := http.NewRequest("POST", "/"+randomDeviceUDID+"/launch?bundleID=com.apple.Preferences", nil)
+	if err != nil {
+		t.Error(err)
+		t.FailNow()
+	}
+	r.ServeHTTP(responseRecorder, launchAppRequest)
+	require.Equal(t, http.StatusBadRequest, responseRecorder.Code, "Code should be BadRequest if X-GO-IOS-RESERVE header is missing")
+
+	var response GenericResponse
+	responseData, _ := ioutil.ReadAll(responseRecorder.Body)
```
danielpaulus, line 140: ioutil.ReadAll is deprecated. Just use io.ReadAll please.

restapi/api/reserve_endpoints_test.go
```diff
@@ -122,9 +124,147 @@ func TestValidateDeviceNotReserved(t *testing.T) {
...
+	err = json.Unmarshal(responseData, &response)
+	if err != nil {
+		t.Error(err)
+		t.FailNow()
+	}
+
+	require.NotEmpty(t, response.Error, "There is no error message returned when X-GO-IOS-RESERVE header is missing")
```
danielpaulus, line 171: Could be good also to verify the correct HTTP Status codes are sent.

danielpaulus (changes requested):
A few minor discussion points. In general it looks pretty good!

restapi/api/middleware.go
```diff
@@ -68,3 +69,21 @@ func StreamingHeaderMiddleware() gin.HandlerFunc {
 		c.Next()
 	}
 }
+
+func ReserveDevicesMiddleware() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		udid := c.Param("udid")
```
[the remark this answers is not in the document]
shamanec: I assumed that all device related endpoints will go through the `DeviceMiddleware` either way, that is why there is no check here

restapi/api/reserve_endpoints.go
```diff
@@ -12,6 +13,7 @@ import (
 var reservedDevicesMap = make(map[string]*reservedDevice)
 var reserveMutex sync.Mutex
 var ReservedDevicesTimeout time.Duration = 5
```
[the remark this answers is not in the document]
shamanec: good catch, forgot about that after changing the packages of the reservation!

restapi/api/routes.go
shamanec: yep, that was on my mind when adding the endpoints but completely forgot about it, good point!

restapi/api/reserve_endpoints_test.go
shamanec: that is actually validated a few lines above, in all tests I check response code first before proceeding with other checks :P

restapi/api/middleware.go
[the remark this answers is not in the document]
danielpaulus: yeah makes sense

restapi/api/reserve_endpoints_test.go
danielpaulus: 🤦  never mind me

restapi/api/routes.go
```diff
@@ -17,22 +18,19 @@ func registerRoutes(router *gin.RouterGroup) {
...
 
 	initAppRoutes(device)
 	initStreamingResponseRoutes(device, router)
-	go reservation.CleanReservationsCRON()
+	go CleanReservationsCRON()
```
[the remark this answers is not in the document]
shamanec: done

restapi/api/routes.go
shamanec: done

restapi/api/reserve_endpoints_test.go
shamanec: done

restapi/api/routes.go
shamanec: I think it makes sense to keep it so we can properly report if the device is no longer connected through the `DeviceMiddleware` and it's also easier to check the map with reserved devices having it. Still let me know if you prefer not having it and I'll rework it

shamanec: Now that I think about it, we shouldn't care if its connected or not, we just want to clear it from the reserved devices list. I'll update it

shamanec: done

danielpaulus (approved):
Thanks for all the work!

PR #199, opened 2022-11-07
Remove panics from code 

grokkos:
Replaced panics with debugging messages and err's

- Todo check the places we need to return the error instead of just info debugging messages and the refactor effort needed in the whole codebase
- Test how the removal of the panics will affect the efficiency and consistency, in extreme occasions that is possibly needed the app to throw panics in order to prevent misbehaviour.

### DIFF
--- a/ios/accessibility/accessibility_control.go
+++ b/ios/accessibility/accessibility_control.go
@@ -1,15 +1,13 @@
 package accessibility
 
 import (
-	"fmt"
-
 	dtx "github.com/danielpaulus/go-ios/ios/dtx_codec"
 	"github.com/danielpaulus/go-ios/ios/nskeyedarchiver"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
-//ControlInterface provides a simple interface to controlling the AX service on the device
-//It only needs the global dtx channel as all AX methods are invoked on it.
+// ControlInterface provides a simple interface to controlling the AX service on the device
+// It only needs the global dtx channel as all AX methods are invoked on it.
 type ControlInterface struct {
 	channel *dtx.Channel
 }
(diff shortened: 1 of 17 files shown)


ios/accessibility/accessibility_control.go
```diff
@@ -19,10 +17,10 @@ func (a ControlInterface) readhostAppStateChanged() {
 		msg := a.channel.ReceiveMethodCall("hostAppStateChanged:")
 		stateChange, err := nskeyedarchiver.Unarchive(msg.Auxiliary.GetArguments()[0].([]byte))
 		if err != nil {
-			panic(err)
+			logrus.WithError(err).Error("Failed unarchiving NSKeyedArchiver plists")
```
danielpaulus, line 20: Can you please change back `logrus` to log? 
this is a little verbose for me :-)

danielpaulus (changes requested):

ios/accessibility/accessibility_control.go
grokkos: sure! done it in all the other places as well, let me know if it looks good or I missed smt :D

restapi/api/util.go
```diff
@@ -31,15 +31,15 @@ func GetVersion() string {
 func MustMarshal(v interface{}) string {
 	b, err := json.Marshal(v)
 	if err != nil {
-		panic(err)
+		log.WithError(err).Error("Failed marshalling object")
```
danielpaulus, line 34: If it's in a `MustXXX` function then throwing a panic is ok. Also the restapi is not library code, so panic'ing here is usually ok.

restapi/api/util.go
```diff
@@ -10,16 +10,16 @@ import (
...
 )
 
 type GenericResponse struct {
 	Message string `json:"message,omitempty"`
 	Error   string `json:"error,omitempty"`
 }
 
-//GetVersion reads the contents of the file version.txt and returns it.
-//If the file cannot be read, it returns "could not read version"
+// GetVersion reads the contents of the file version.txt and returns it.
```
danielpaulus, line 21: While I don't have a problem with adding these whitespaces in general, it usually is a good practice not to mix code cleanup/automated refactoring with your changes. It makes it hard to review and is a risk of adding merge conflicts. Please remove them from this PR, I am ok with merging them separately.

danielpaulus (changes requested):

PR #208, opened 2022-11-27
Add pairing endpoint to REST API

shamanec:
* Added a new endpoint for pairing (un)supervised devices  
* Updated Swagger generation on the device endpoints a little bit


restapi/api/device_endpoints.go
```diff
@@ -293,3 +302,57 @@ func DisableDeviceCondition(c *gin.Context) {
...
+
+	file, _, err := c.Request.FormFile("p12file")
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, GenericResponse{Error: "Could not parse p12 file from form-data, err:" + err.Error()})
+		return
+	}
+	p12fileBuf := new(bytes.Buffer)
+	p12fileBuf.ReadFrom(file)
+
+	supervision_password := c.Request.FormValue("supervision_password")
```
danielpaulus, line 349: could also go in as a header param. maybe nicer than a form parameter

danielpaulus (changes requested):
wdyt about moving the password to a header?

restapi/api/device_endpoints.go
shamanec: I was wondering which path to take, this is not production ready so security is not an issue xD having it in a header is as good as any, I'll change it

danielpaulus: It's less security, more UX. I think putting it in the header might be easier to do because it already has key value format. and than you can just post the file which should simplify caller code.

shamanec: moved password to a header, instead of part of form-data

danielpaulus (approved):
Good stuff! Thanks for updating the docs. Good docs will be 80% of the work here imo.

PR #221, opened 2023-01-10
Add --r to fsync rm

YasharAzari:
Add a new `RemoveAll` function to `fsync` and a new flag to `fsync rm` for recursive deletion from a target path


danielpaulus (approved):
Nice addition! Thanks!

PR #271, opened 2023-07-23
add example config, a testcase to use in the debugger and decoder cod…

danielpaulus:
…e for XCTTestIdentifiers


danielpaulus:
Closing — this was an unfinished draft (leftover debug `print` calls, swallowed errors). The XCTTestIdentifierSet decoding could be salvaged later but needs cleanup before it's mergeable.

PR #280, opened 2023-09-03
`fsync`: support app container jails

mhils:
First off, thanks for the fantastic work on go-ios! :smiley: 🍰 Extending go-ios was super easy after figuring out what I needed with the debug proxy.

I've had a use case where I wanted to read/write files from/to a specific app container jail. Fortunately `house_arrest` already had everything I needed, and adding an `--app` switch to the fsync commands was straightforward.

1. Is accessing app jails for fsync something you folks are interested in supporting?
2. If yes, where do we want vendContainer to live? I've just copied it over from `house_arrest.go` to avoid a circular dependency, but I'm happy to do the legwork and move it into something like `afc/house_arrest.go` and have everything point there.

### DIFF
--- a/ios/afc/fsync.go
+++ b/ios/afc/fsync.go
@@ -3,6 +3,7 @@ package afc
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/danielpaulus/go-ios/ios"
 	log "github.com/sirupsen/logrus"
+	"howett.net/plist"
 )
 
 const serviceName = "com.apple.afc"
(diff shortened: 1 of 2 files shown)


ios/afc/fsync.go
```diff
@@ -48,6 +50,68 @@ func New(device ios.DeviceEntry) (*Connection, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	err = vendContainer(deviceConn, bundleID)
+	if err != nil {
+		return nil, err
+	}
+	return &Connection{deviceConn: deviceConn}, nil
+}
+
```
mhils, line 64: ⬇️  all this is simply copied from `house_arrest.go`

danielpaulus:
@mhils Thanks for this addition to the project!
I took a look at the modules right now and it is indeed a bit messy the way it is. 
house_arrest.go just duplicates most code that is already in fsync.go. If you can do this, I would appreciate if you can just move the entire code over to the afc package. 
The only thing that should stay in house_arrest.go are is the public interface for compatibility reasons. 
So, 
New()
Close()
SendFile()
ListFiles()
CloseFile()
should be the only functions to stay but use afc functions.

aluedeke:
Those chan ges have been merged as part of #555

PR #284, opened 2023-09-25
Adding support to modify Time Format

harshitshah4:
Added support to modify timeformat of an iOS device.
In lockdown, there is a key `Uses24HourClock` that returns values as true or false, when 24 Hour Clock is enabled or disabled, respectively.
Turns out this key's value can also be modified over Lockdown Connection, to enable or disable 24 Hour Clock on the device, in turn changing time format of the device.

image

### DIFF
--- a/main.go
+++ b/main.go
@@ -1346,7 +1346,7 @@ func timeFormat(device ios.DeviceEntry, operation string, force bool) {
 			timeFormat = "12h"
 		}
 		if JSONdisabled {
-			fmt.Printf("%t\n", timeFormat)
+			fmt.Printf("%s\n", timeFormat)
 		} else {
 			fmt.Println(convertToJSONString(map[string]string{"TimeFormat": timeFormat}))
 		}
--- a/main.go
+++ b/main.go
@@ -119,6 +119,7 @@ Usage:
   ios assistivetouch (enable | disable | toggle | get) [--force] [options]
   ios voiceover (enable | disable | toggle | get) [--force] [options]
   ios zoomtouch (enable | disable | toggle | get) [--force] [options]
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -215,7 +216,8 @@ The commands work as following:
    ios resetlocation [options]                                        Resets the location of the device to the actual one
    ios assistivetouch (enable | disable | toggle | get) [--force] [options] Enables, disables, toggles, or returns the state of the "AssistiveTouch" software home-screen button. iOS 11+ only (Use --force to try on older versions).
    ios voiceover (enable | disable | toggle | get) [--force] [options] Enables, disables, toggles, or returns the state of the "VoiceOver" software home-screen button. iOS 11+ only (Use --force to try on older versions).
-   ios zoom (enable | disable | toggle | get) [--force] [options] Enables, disables, toggles, or returns the state of the "ZoomTouch" software home-screen button. iOS 11+ only (Use --force to try on older versions). 
+   ios zoom (enable | disable | toggle | get) [--force] [options] Enables, disables, toggles, or returns the state of the "ZoomTouch" software home-screen button. iOS 11+ only (Use --force to try on older versions).
+   ios timeformat (24h | 12h | get) [--force] [options] Sets, or returns the state of the "time format". iOS 11+ only (Use --force to try on older versions).
```
shamanec, line 220: You've missed the toggle option here:
```suggestion
   ios timeformat (24h | 12h | toggle | get) [--force] [options] Sets, or returns the state of the "time format". iOS 11+ only (Use --force to try on older versions).
```

shamanec:
lgtm

shamanec:
Thanks for the contribution!

harshitshah4:
Thanks alot @shamanec for quick review and all the help.

PR #297, opened 2023-10-27
fix: cannot find path to paste in postinstall.js

IsLand-x:


danielpaulus:
I removed this command because it does not work with never NPM versions. F.ex. npm 9.6.7  fails with "Unknown command: 'bin'"
Why do you need this back?

danielpaulus:
Thanks for this! Unfortunately `npm bin -g` was removed in npm 9+, so re-enabling it would break installs on modern npm. The install path is now resolved via `npm_config_prefix` (plus a Windows-specific fix in #730), so this is no longer needed. Closing.

PR #298, opened 2023-10-30
Draft: Reusable XPC Connection type

diegoperini:


ios/xpc/encoding.go
```diff
@@ -111,6 +112,28 @@ func EncodeData(w io.Writer, body map[string]interface{}) error {
 	return err
 }
 
+// TODO : remove this and figure out how to reuse EncodeData() to simulate what happens under this function
+func EncodeEmpty(w io.Writer, messageId uint64, additionalXpcFlags uint32, initHandshake bool) error {
```
dmissmann, line 116: You could add a function `Encode(w io.Writer, m Message) error` (and also add `Id uint64` as a field to `Message`). Trying to derive the values for `Message` inside `EncodeData` doesn't make much sense anyways, the encoding part should only care about encoding

ios/xpc/connection.go
```diff
@@ -0,0 +1,166 @@
...
+		if firstReadFrame.Header().Type != http2.FrameSettings {
+			return nil, errors.New("Received unexpected frame from XPC connection")
+		}
+		// TODO : figure out if need to act on this frame
+	}
+
+	err = framer.WriteSettingsAck()
+	if err != nil {
+		return nil, err
+	}
```
dmissmann, line 81: For readability this can be extracted into it's own method

ios/xpc/connection.go
```diff
@@ -0,0 +1,166 @@
...
+	_, err = conn.Receive() // TODO : figure out if need to act on this frame
+	if err != nil {
+		return nil, err
+	}
+	_, err = conn.Receive() // TODO : figure out if need to act on this frame
+	if err != nil {
+		return nil, err
+	}
```
dmissmann, line 125: This can be extracted as well. And since we have three pairs of send/receive I'm pretty sure the receive part can be done after each send.

ios/xpc/connection.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return conn, nil
+}
+
+func (c *Connection) Receive() (map[string]interface{}, error) {
+	for true {
```
dmissmann, line 131:
```suggestion
	for {
```

ios/xpc/connection.go
```diff
@@ -0,0 +1,166 @@
...
+			msg, err := DecodeMessage(bytes.NewReader(dataFrame.Data()))
+			if err != nil {
+				return nil, err
+			}
+			return msg.Body, nil
+		}
+	}
+
+	// Impossible to reach here
+	return map[string]interface{}{}, nil
```
dmissmann, line 151: If it's not reachable you can remove it 😉 (the compiler only complains because you wrote `for true {`  instead of `for {`)

ios/xpc/connection.go
```diff
@@ -0,0 +1,166 @@
...
+		response, err := c.framer.ReadFrame()
+		if err != nil {
+			return nil, err
+		}
+
+		if response.Header().Type == http2.FrameData {
+			dataFrame := response.(*http2.DataFrame)
+			if dataFrame.Length == 24 { // empty payload
+				return nil, nil
+			}
```
dmissmann, line 141: Lets still run the decoder, otherwise you hide a potential error when decoding fails for some reason

ios/xpc/encoding.go
diegoperini: Done, there is only one `EncodeMessage` function now.

ios/xpc/connection.go
diegoperini: Done

ios/xpc/connection.go
diegoperini: Removed. 🙇

ios/xpc/connection.go
diegoperini: Done

dmissmann (approved):

PR #302, opened 2023-11-10
WIP: Find IP based on packets without MAC

eyJhb:
This implements a way to find the IP address of the device, without using the mac address of the device.
Basically just counts the number of occurrences of the IP in the src, and then takes the one with the most, as it's most likely to be the correct one.

Works fine, if you make some noise while it's running.

I can't get the `--duration` to work however, not sure why.

But I would appreciate some comments on this.

Also functions are not named `findIp`, `findIp` and `findIP` anymore. Should be more readable. 

@shamanec and @danielpaulus WIP as I would appreciate any comments on this.


danielpaulus:
@eyJhb  haha that's a clever hack! I like it :-D 
Probabilistic IP analysis. I'll take a look now.

ios/pcap/ipfinder.go
```diff
@@ -13,81 +16,172 @@ type NetworkInfo struct {
...
+			return NetworkInfo{Mac: packet.Mac, IPv4: packet.IPv4, IPv6: packet.IPv6}, nil
+		}
+	}
+
+	return NetworkInfo{}, nil
 }
 
-func findIp(device ios.DeviceEntry, mac string) (NetworkInfo, error) {
-	intf, err := ios.ConnectToService(device, "com.apple.pcapd")
+// FindIp reads pcap packets until one is found that matches the given MAC
```
danielpaulus, line 61: please add a comment explaining how this works

ios/pcap/ipfinder.go
```diff
@@ -13,81 +16,172 @@ type NetworkInfo struct {
...
+		}
+	}
+
+	return NetworkInfo{IPv4: highestIPv4Addr, IPv6: highestIPv6Addr}, nil
+}
+
+func startCapture(device ios.DeviceEntry, c chan PacketInfo, done chan bool) error {
+	intf, err := ios.ConnectToService(device, "com.apple.pcapd")
+	if err != nil {
+		return err
```
danielpaulus, line 121: I recently learned that for debugging things later, in Golang it is a good practice to wrap errors and indicate the function name of where they happened. So rather than just returning err here and in other places, it is better to do sth like: 
return fmt.Errorf("startCapture: failed connecting to com.apple.pcapd with err: %w, err)

eyJhb:
> @eyJhb haha that's a clever hack! I like it :-D Probabilistic IP analysis. I'll take a look now.
> 
> wdyt about adding a flag that automatically creates noise? I found that opening Safari, or the weather app usually causes network traffic. Maybe you can find other things too. We could offer this as part of this implementation.

Thanks! I quite like it as well! :D - We could just open Apple Maps :) That generates a lot of traffic, and was something I was playing with. It could be a flag to disable/enable it.

I'll be looking at the PR again after Christmas, hopefully that's OK with you!

danielpaulus:
> > @eyJhb haha that's a clever hack! I like it :-D Probabilistic IP analysis. I'll take a look now.
> > wdyt about adding a flag that automatically creates noise? I found that opening Safari, or the weather app usually causes network traffic. Maybe you can find other things too. We could offer this as part of this implementation.
> 
> Thanks! I quite like it as well! :D - We could just open Apple Maps :) That generates a lot of traffic, and was something I was playing with. It could be a flag to disable/enable it.
> 
> I'll be looking at the PR again after Christmas, hopefully that's OK with you!

That's a really good idea :-D 
Yeah of course, whenever you have the time. I am grateful for every contribution. Enjoy your Christmas vacation then 👍

ios/pcap/ipfinder.go
eyJhb: Should be resolved in latest commit.

ios/pcap/ipfinder.go
eyJhb: Should be resolved in latest commit.

eyJhb:
So I've fixed the two comments, but I haven't added another flag. I've thought if it wouldn't make more sense, to just tell the user to run something like `ios launch <name-of-apple-maps-that-i-cant-remember> && ios ip --lazy`, instead. This reduces the complexity of the implementation a lot :) What are your thoughts on this?

Also, I still have _no_ clue why `duration` doesn't want to work from the arguments. Can you aid in that @danielpaulus ?

danielpaulus:
I'll take a look later today. I love this PR, it wins my prize of most creative solution ever :-D

danielpaulus:
I changed it to this to remove the goroutines 
```// FindIPByLazy reads pcap packets for a specified duration, whereafter it will
// try to find the IP address that occurs the most times, and assume that is the IP of the device.
// This is of course based on, that the device contacts mulitple IPs, and that there is some traffic.
// If the device only contains a single IP, then it would be 50/50 which IP will be returned.
// This is best effort! It's important to generate some traffic, when this function runs to get better results.
func FindIPByLazy(device ios.DeviceEntry, capture_duration time.Duration) (NetworkInfo, error) {
	pcapService, err := ios.ConnectToService(device, "com.apple.pcapd")
	if err != nil {
		return NetworkInfo{}, fmt.Errorf("FindIPByLazy: failed connecting to com.apple.pcapd with err: %w", err)
	}

	ipv6Hits := make(map[string]int)
	ipv4Hits := make(map[string]int)

	select {
	case <-time.After(capture_duration):
		break
	default:
		packet, err := readPacket(pcapService.Reader())
		if err != nil {
			return NetworkInfo{}, fmt.Errorf("FindIPByLazy: error reading pcap packet: %w", err)
		}
		if packet.IPv4 != "" {
			ipv4Hits[packet.IPv4] += 1
		}
		if packet.IPv6 != "" {
			ipv6Hits[packet.IPv6] += 1
		}

	}

	highestIPv4Hits, highestIPv4Addr := 0, ""
	for ipv4, hits := range ipv4Hits {
		if hits > highestIPv4Hits {
			highestIPv4Hits = hits
			highestIPv4Addr = ipv4
		}
	}

	highestIPv6Hits, highestIPv6Addr := 0, ""
	for ipv6, hits := range ipv6Hits {
		if hits > highestIPv6Hits {
			highestIPv6Hits = hits
			highestIPv6Addr = ipv6
		}
	}

	return NetworkInfo{IPv4: highestIPv4Addr, IPv6: highestIPv6Addr}, nil
}

var plistCodec = ios.NewPlistCodec()

func readPacket(r io.Reader) (PacketInfo, error) {
	b, err := plistCodec.Decode(r)
	if err != nil {
		return PacketInfo{}, fmt.Errorf("readPacket: failed decoding plistCodec err: %w", err)
	}
	decodedBytes, err := fromBytes(b)
	if err != nil {
		return PacketInfo{}, fmt.Errorf("readPacket: failed decoding fromBytes err: %w", err)
	}
	_, packet, err := getPacket(decodedBytes)
	if err != nil {
		return PacketInfo{}, fmt.Errorf("readPacket: failed getPacket err: %w", err)
	}
	if len(packet) > 0 {
		pInfo := parsePacket(packet)
		if pInfo.complete() {
			return pInfo, nil
		}
	}
	return PacketInfo{}, nil
}
```
Wdyt?

danielpaulus:
main.go I did like this: 
```
ios ip [options] [--lazy=<duration_seconds>]
```

and then:

```
	b, _ = arguments.Bool("ip")
	if b {
		var ip pcap.NetworkInfo
		var err error

		lazy, _ := arguments.String("--lazy")
		if lazy == "" {
			ip, err = pcap.FindIPByMac(device)
		} else {
			seconds, err := arguments.Int("--lazy")
			if seconds <= 0 {
				log.Fatal("Please provide a positive, non zero integer for the duration.")
			}
			if err != nil {
				exitIfError("please provide an integer for the duration. like 'ios ip --lazy=5' for 5 seconds", err)
			}
			ip, err = pcap.FindIPByLazy(device, time.Duration(seconds)*time.Second)
			exitIfError("failed finding lazy IP", err)

		}
		exitIfError("failed", err)
		println(convertToJSONString(ip))
		return
	}
```

eyJhb:
> I'll take a look later today. I love this PR, it wins my prize of most creative solution ever :-D

Thanks! It's a neat little hack, when Apple doesn't want to expose such information (weirdos) :D 

> I changed it to this to remove the goroutines
> ...
> Wdyt?

Much cleaner! Would much rather have that, than the goroutines! I'll be updating the PR to reflect that change. :)



> main.go I did like this:
> 
> ...
> 
> and then:
> 
> ...

Perfect, thank you. I'll be playing around with that! :)
Hopefully I can get `--lazy` to behave without any duration added, and default to 30s.

Hopefully I'll be able to have it all put together, and tested before the end of the day.

eyJhb:
So I've updated the PR with the latest changes. I really like the idea of using the timer for stopping the process, great work!
I've added a timeout for finding IP by mac as well, as it made sense to have in both. The syntax is now

`go-ios ip --lazy --timeout=30`
`go-ios ip --timeout=30`

It defaults to 10s timeout., which should be fine?

Sorry for not doing this before, it's been quite a busy time....

eyJhb:
I've squashed my commits into a single one, I think this is good to merge? Please test as well to ensure it however.

danielpaulus:
sorry for the delay @eyJhb  we've been a bit busy with the ios17 work. I will check and merge this today. Thank you so much for this clever solution :-D

eyJhb:
> sorry for the delay @eyJhb we've been a bit busy with the ios17 work. I will check and merge this today. Thank you so much for this clever solution :-D

Awesome! Take your time, I think that ios17 work is much more important :) So no pressure from me. If there is anything I can help you with to speed up the process, please do let me know! :)

eyJhb:
@danielpaulus any time for reviewing this? :D

danielpaulus:
Thank you for this, @eyJhb 🙏 — and apologies it stayed open as a WIP for so long. The motivation is real and increasingly important: iOS randomizes Wi-Fi MACs, so MAC-based IP discovery is fragile, and a MAC-independent path is worth having. The timeout you added to the existing MAC search is also a legitimate fix for a real hang.

I'm closing it because it's still WIP and the base has moved a long way:
- The 'most frequent source IP' heuristic tallies remote packets too (a busy APNs/iCloud/CDN peer or a download dominates), so it can return a non-device IP — it needs direction/interface metadata or private-range correlation to be reliable.
- Since the Nov-2023 base, logging migrated to `golog` and `ios ip` moved from `main.go` into `cmd_device_diagnostics.go` + the command registry, so the diff no longer applies; it also renames the exported `FindIp` without updating callers.

If you're still interested, a fresh PR on current main would be very welcome: keep the `--lazy`/`--timeout` flag shape, but base the discovery on pcapd direction/interface info (or RSD-first) rather than raw frequency, keep `FindIp` API-compatible, and add unit tests over canned packet bytes. Happy to help scope it. Thanks again for pushing on this — the idea is a good one. 💙

PR #307, opened 2023-11-29
Replace DeviceConnectionInterface with io.ReadWriteCloser for DTX

dmissmann:
The `DeviceConnectionInterface` doesn't look like a very useful interface. Most of the service should only require a `io.ReadWriteCloser` to communicate with the device. And that's what kind of already happens, but since `DeviceConnectionInterface` exposes the `io.Reader` and `io.Writer` interfaces on separate methods, we still have to go through that instead of simply using the `io.ReadWriteCloser` interface directly.

For DTX the changes are very minimal, and most other services should not be more complex as most of the additional methods in `DeviceConnectionInterface` are either used by the debugproxy or lockdown.

### DIFF
--- a/ios/deviceconnection.go
+++ b/ios/deviceconnection.go
@@ -13,9 +13,11 @@ import (
 
 // DeviceConnectionInterface contains a physical network connection to a usbmuxd socket.
 type DeviceConnectionInterface interface {
-	Close() error
+	// Deprecated: use the Read([]byte) (int, error) function instead
 	Send(message []byte) error
+	// Deprecated: use the Read([]byte) (int, error) function
 	Reader() io.Reader
+	// Deprecated: use the Write([]byte) (int, error) function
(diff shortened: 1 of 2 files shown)


danielpaulus:
> The `DeviceConnectionInterface` doesn't look like a very useful interface. Most of the service should only require a `io.ReadWriteCloser` to communicate with the device. And that's what kind of already happens, but since `DeviceConnectionInterface` exposes the `io.Reader` and `io.Writer` interfaces on separate methods, we still have to go through that instead of simply using the `io.ReadWriteCloser` interface directly.
> 
> For DTX the changes are very minimal, and most other services should not be more complex as most of the additional methods in `DeviceConnectionInterface` are either used by the debugproxy or lockdown.

"Doesn't look very useful".. you are really hurting my feelings here @dmissmann  😭 
🤣  I think the change makes sense.

PR #308, opened 2023-12-11
Mount personalized developer disk images

dmissmann:
With iOS 17 we don't have developer disks per minor version that we can simply mount, we get "personalized" images where we have to ask apple for a signature first.

The cli command itself hasn't changed much, only the path to the image is now different. You need to mount the image at `~/Library/Developer/DeveloperDiskImages/iOS_DDI.dmg`, and then the path to `go-ios` should be like `/Volumes/Xcode_iOS_DDI_Personalized/Restore`. Pointing directly to `~/Library/Developer/DeveloperDiskImages/iOS_DDI.dmg` was not implemented as `dmg` files are not supported by the go std lib, and therefore to keep the implementation simpler.

### DIFF
--- a/ios/imagemounter/personalized_image_mounter.go
+++ b/ios/imagemounter/personalized_image_mounter.go
@@ -26,7 +26,7 @@ type PersonalizedDeveloperDiskImageMounter struct {
 func NewPersonalizedDeveloperDiskImageMounter(entry ios.DeviceEntry, version *semver.Version) (PersonalizedDeveloperDiskImageMounter, error) {
 	values, err := ios.GetValuesPlist(entry)
 	if err != nil {
-		return PersonalizedDeveloperDiskImageMounter{}, nil
+		return PersonalizedDeveloperDiskImageMounter{}, fmt.Errorf("NewPersonalizedDeveloperDiskImageMounter: could not read lockdown values: %w", err)
 	}
 	var ecid uint64
 	if e, ok := values["UniqueChipID"].(uint64); ok {
--- a/ios/imagemounter/personalized_image_mounter.go
+++ b/ios/imagemounter/personalized_image_mounter.go
@@ -202,7 +202,7 @@ func (p PersonalizedDeveloperDiskImageMounter) mountPersonalizedImage(signatureB
 func getFileSize(p string) (uint64, error) {
 	info, err := os.Stat(p)
 	if err != nil {
-		return 0, err
+		return 0, fmt.Errorf("getFileSize: could not get file stats for '%s': %w", p, err)
 	}
 	if info.IsDir() {
 		return 0, fmt.Errorf("getFileSize: expected a file, but got a directory: '%s'", p)
--- a/ios/imagemounter/personalized_image_mounter.go
+++ b/ios/imagemounter/personalized_image_mounter.go
@@ -1,6 +1,7 @@
 package imagemounter
 
 import (
+	"fmt"
 	"github.com/Masterminds/semver"
 	"github.com/danielpaulus/go-ios/ios"
 	log "github.com/sirupsen/logrus"
@@ -12,6 +13,7 @@ type personalizedDeveloperDiskImageMounter struct {
 	plistRw    ios.PlistCodecReadWriter
 	version    *semver.Version
 	tss        tssClient
+	ecid       uint64
 }
 
 func (p personalizedDeveloperDiskImageMounter) ListImages() ([][]byte, error) {
@@ -20,7 +22,28 @@ func (p personalizedDeveloperDiskImageMounter) ListImages() ([][]byte, error) {
 
 func (p personalizedDeveloperDiskImageMounter) MountImage(imagePath string) error {
 	manifest, err := loadBuildManifest(path.Join(imagePath, "BuildManifest.plist"))
-	//TODO implement me
+	if err != nil {
+		return fmt.Errorf("failed to load build manifest. %w", err)
+	}
+
+	identifiers, err := p.queryIdentifiers()
+	if err != nil {
+		return fmt.Errorf("failed to query personalization identifiers. %w", err)
+	}
+	nonce, err := p.queryPersonalizedImageNonce()
+	if err != nil {
+		return fmt.Errorf("failed to get nonce. %w", err)
+	}
+
+	identity, err := manifest.findIdentity(identifiers)
+	if err != nil {
+		return err
+	}
+
+	signature, err := p.tss.getSignature(identity, identifiers, nonce, p.ecid)
+	if err != nil {
+		return fmt.Errorf("failed to get signature from Apple. %w", err)
+	}
 	panic("implement me")
 }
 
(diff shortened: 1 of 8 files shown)


ios/imagemounter/personalized_image_mounter.go
```diff
@@ -0,0 +1,196 @@
...
+	plistRw    ios.PlistCodecReadWriter
+	version    *semver.Version
+	tss        tssClient
+	ecid       uint64
+}
+
+func NewPersonalizedDeveloperDiskImageMounter(entry ios.DeviceEntry, version *semver.Version) (PersonalizedDeveloperDiskImageMounter, error) {
+	values, err := ios.GetValuesPlist(entry)
+	if err != nil {
+		return PersonalizedDeveloperDiskImageMounter{}, nil
```
danielpaulus, line 24: @dmissmann you are swallowing the error here, please wrap it like fmt.Errorf("NewPersonalizedDeveloperDiskImageMounter: failed to get lockdown values from device with error: %w", err)

ios/imagemounter/personalized_image_mounter.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	var ecid uint64
+	if e, ok := values["UniqueChipID"].(uint64); ok {
+		ecid = e
+	} else {
+		return PersonalizedDeveloperDiskImageMounter{}, fmt.Errorf("could not get ECID from device")
```
danielpaulus, line 30: I want to start adding function names to errors so it will be easier to debug later. 
Can you add "NewPersonalizedDeveloperDiskImageMounter:" in front of the error please?

ios/imagemounter/personalized_image_mounter.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	deviceConn, err := ios.ConnectToService(entry, serviceName)
+	if err != nil {
+		return PersonalizedDeveloperDiskImageMounter{}, err
```
danielpaulus, line 34: same here, please start wrapping errors like above. I recently learned that this is how you do it so we can later properly debug things when they fail. 
fmt.Errorf("NewPersonalizedDeveloperDiskImageMounter: failed connecting to serviceName with err: %w", err)

ios/imagemounter/personalized_image_mounter.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (p PersonalizedDeveloperDiskImageMounter) ListImages() ([][]byte, error) {
+	return listImages(p.plistRw, "Personalized", p.version)
+}
+
+func (p PersonalizedDeveloperDiskImageMounter) MountImage(imagePath string) error {
+	manifest, err := loadBuildManifest(path.Join(imagePath, "BuildManifest.plist"))
+	if err != nil {
+		return fmt.Errorf("failed to load build manifest. %w", err)
```
danielpaulus, line 56: please add the functionname to the error MountImage:

ios/imagemounter/personalized_image_mounter.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	identifiers, err := p.queryIdentifiers()
+	if err != nil {
+		return fmt.Errorf("failed to query personalization identifiers. %w", err)
```
danielpaulus, line 61: same here and all errors below, I won'T add comments for each one. Please check it yourself.

danielpaulus (changes requested):
I recently learned that, for being able to actually debug golang errors later on, you need to somewhat "build the stacktrace" you get with Java yourself. 
So, errors should always be wrapped and include the func name where they happen. So that when we later get bug reports from users, we can pinpoint what caused the errors easier.

ios/plistcodec.go
```diff
@@ -56,3 +57,50 @@ func (plistCodec PlistCodec) Decode(r io.Reader) ([]byte, error) {
 	}
 	return payloadBytes, nil
 }
+
+type PlistCodecReadWriter struct {
```
danielpaulus, line 61: I think this is a really nice idea! 
Please add comments explaining why it's needed. 
Also comment public methods please.

ios/plistcodec.go
```diff
@@ -56,3 +57,50 @@ func (plistCodec PlistCodec) Decode(r io.Reader) ([]byte, error) {
...
+func (p PlistCodecReadWriter) Write(m interface{}) error {
+	stringContent := ToPlist(m)
+	log.Tracef("Lockdown send %v", reflect.TypeOf(m))
+	buf := new(bytes.Buffer)
+	length := len(stringContent)
+	messageLength := uint32(length)
+
+	err := binary.Write(buf, binary.BigEndian, messageLength)
+	if err != nil {
+		return err
```
danielpaulus, line 82: please wrap the error with the struct and func name, like
PlistCodecReadWriter.Write:

ios/plistcodec.go
```diff
@@ -56,3 +57,50 @@ func (plistCodec PlistCodec) Decode(r io.Reader) ([]byte, error) {
...
+	}
+	length := binary.BigEndian.Uint32(buf)
+	payloadBytes := make([]byte, length)
+	n, err := io.ReadFull(p.r, payloadBytes)
+	if err != nil {
+		return fmt.Errorf("lockdown Payload had incorrect size: %d expected: %d original error: %s", n, length, err)
+	}
+	_, err = plist.Unmarshal(payloadBytes, v)
+	return err
```
danielpaulus, line 105: als this errror and the others above, please wrap 'em

ios/imagemounter/personalized_image_mounter.go
dmissmann: sure, how do you want to handle function names that are present in many places like `Write`? Only the function name, or also the type if available?
I guess in most cases the name+message should be enough to get to the right place easily

dmissmann:
I think I handled all the error messages now with the additional name in them, but only for new code. Would do it for the old stuff as well, but in a separate PR

ios/imagemounter/personalized_image_mounter.go
danielpaulus: Good question, I think yeah the function name is probably enough. Also if we do this consistently from now on, you should be able to get something like a Java Stacktrace, so it should be relatively clear which call to Write it was from the other errrors wrapping this one.

main.go
```diff
@@ -142,6 +142,7 @@ The commands work as following:
    ios info [options]                                                 Prints a dump of Lockdown getValues.
    ios image list [options]                                           List currently mounted developers images' signatures
    ios image mount [--path=<imagepath>] [options]                     Mount a image from <imagepath>
+   >                                                                  For iOS 17+ (personalized deverloper disk images) <imagepath> must point to the "Restore" directory inside the developer disk
```
danielpaulus, line 145: typo: deverloper

danielpaulus (changes requested):
This is really excellent work! Thanks a lot for this :-)
There is just s small typo in the main.go, otherwise it is good to go

danielpaulus:
> I think I handled all the error messages now with the additional name in them, but only for new code. Would do it for the old stuff as well, but in a separate PR

Thank you! 
That is something that we indeed can and should do later, as part of regular clean up and refactoring. Getting iOS17 working has higher priority though.

danielpaulus (approved):

PR #321, opened 2023-12-26
Developer mode

shamanec:
Introduce capability to check current developer mode status or enable developer mode on a device  
All credit goes to [pymobiledevice3](https://github.com/doronz88/pymobiledevice3) and doronz88 for finding the way to do this

### DIFF
--- a/ios/amfi/amfi.go
+++ b/ios/amfi/amfi.go
@@ -1,0 +1,100 @@
+package amfi
+
+import (
+	"fmt"
+
+	"github.com/danielpaulus/go-ios/ios"
+)
+
+const serviceName string = "com.apple.amfi.lockdown"
+
+type Connection struct {
+	deviceConn ios.DeviceConnectionInterface
+	plistCodec ios.PlistCodec
+}
+
+func New(device ios.DeviceEntry) (*Connection, error) {
+	deviceConn, err := ios.ConnectToService(device, serviceName)
+	if err != nil {
+		return &Connection{}, err
+	}
+
+	var devModeConn Connection
+	devModeConn.deviceConn = deviceConn
+	devModeConn.plistCodec = ios.NewPlistCodec()
+
+	return &devModeConn, nil
+}
+
+func (devModeConn *Connection) Close() error {
+	return devModeConn.deviceConn.Close()
+}
+
+func (devModeConn *Connection) EnableDevMode() error {
+	reader := devModeConn.deviceConn.Reader()
+
+	request := map[string]interface{}{"action": 1}
+
+	bytes, err := devModeConn.plistCodec.Encode(request)
+	if err != nil {
+		return err
+	}
+
+	err = devModeConn.deviceConn.Send(bytes)
+	if err != nil {
+		return err
+	}
+
+	responseBytes, err := devModeConn.plistCodec.Decode(reader)
+	if err != nil {
+		return err
+	}
+
+	plist, err := ios.ParsePlist(responseBytes)
+	if err != nil {
+		return err
+	}
+
+	// Check if we have an error returned by the service
+	if _, ok := plist["Error"]; ok {
+		return fmt.Errorf("amfi could not enable developer mode")
+	}
+
+	if _, ok := plist["success"]; ok {
+		return nil
+	}
+
+	return fmt.Errorf("amfi could not enable developer mode but no error or success was reported")
+}
+
+func (devModeConn *Connection) EnableDevModePostRestart() error {
+	reader := devModeConn.deviceConn.Reader()
+
+	request := map[string]interface{}{"action": 2}
+
+	bytes, err := devModeConn.plistCodec.Encode(request)
+	if err != nil {
+		return err
+	}
+
+	err = devModeConn.deviceConn.Send(bytes)
+	if err != nil {
+		return err
+	}
+
+	responseBytes, err := devModeConn.plistCodec.Decode(reader)
+	if err != nil {
+		return err
+	}
+
+	plist, err := ios.ParsePlist(responseBytes)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := plist["success"]; ok {
+		return nil
+	}
+
+	return fmt.Errorf("amfi could not enable developer mode post restart")
+}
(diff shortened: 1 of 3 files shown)


ios/amfi/amfi.go
```diff
@@ -0,0 +1,103 @@
...
+	request := map[string]interface{}{"action": 1}
+
+	bytes, err := devModeConn.plistCodec.Encode(request)
+	if err != nil {
+		return fmt.Errorf("EnableDevMode: failed encoding request to service with err: %w", err)
+	}
+
+	err = devModeConn.deviceConn.Send(bytes)
+	if err != nil {
+		return fmt.Errorf("EnableDevMode: failed sending request bytes to service with err: %w", err)
```
danielpaulus, line 46: nice error handling! ❤️

main.go
```diff
@@ -221,6 +224,7 @@ The commands work as following:
    ios timeformat (24h | 12h | toggle | get) [--force] [options] Sets, or returns the state of the "time format". iOS 11+ only (Use --force to try on older versions).
    ios diskspace [options]											  Prints disk space info.
    ios batterycheck [options]                                         Prints battery info.
+   ios dev-mode (enable | get) [--enable-post-restart] [options]	  Enable developer mode on the device or check if it is enabled. Can also completely finalize developer mode setup after device is restarted.
```
danielpaulus, line 227: please change `dev-mode` to `devmode` 
no other command has a dash and I prefer stuff to be consistent

main.go
```diff
@@ -1841,3 +1862,61 @@ func exitIfError(msg string, err error) {
...
+	go func() {
+		time.Sleep(60 * time.Second)
+		done <- true
+	}()
+
+	// Loop trying to reinit the device to find out if it restarted
+WaitLoop:
+	for {
+		select {
+		case <-ticker.C:
```
danielpaulus, line 1898: I think you could also add a <-time.After(60s) as a case here and remove the extra go routine?

main.go
```diff
@@ -1841,3 +1862,61 @@ func exitIfError(msg string, err error) {
 		log.WithFields(log.Fields{"err": err}).Fatalf(msg)
 	}
 }
+
+func enableDevMode(device ios.DeviceEntry, enablePostRestart bool) {
```
danielpaulus, line 1866: I wonder if this command should go to the module and not be in main.go
Could it be useful for others too? What do you think?

danielpaulus (changes requested):
Cool change! Thanks :-)

main.go
shamanec: Yeah, good point, I forgot about that 😅

main.go
shamanec: Mmm I guess people would want to wrap it their way but probably it will be useful for it to be already there. I'll see if I can move it and leave it as cool :D

danielpaulus: maybe provide both? low level control and the high level one here

main.go
shamanec: Done 👌

main.go
shamanec: Done 👌

main.go
shamanec: I moved it completely to the amfi package, just calling it and getting errors in main.go 👌

danielpaulus (approved):

ios/amfi/amfi.go
```diff
@@ -0,0 +1,174 @@
...
+
+	err = conn.EnableDevMode()
+	if err != nil {
+		return fmt.Errorf("EnableDeveloperMode: failed enabling developer mode with err: %w", err)
+	}
+	log.Infof("Successfully enabled developer mode on device `%s`, device will restart", device.Properties.SerialNumber)
+
+	udid := device.Properties.SerialNumber
+	log.Info("Waiting for device to restart after enabling developer mode")
+	ticker := time.NewTicker(10 * time.Second)
```
dmissmann, line 135: 10s per default seems pretty long. And it wouldn't hurt to do this more frequently, the only thing that happens is the `device, err = ios.GetDevice(udid)`, which is pretty cheap.

ios/amfi/amfi.go
```diff
@@ -0,0 +1,174 @@
...
+			if err != nil {
+				log.Info("Device is not yet available")
+				continue WaitLoop
+			}
+			break WaitLoop
+		case <-time.After(60 * time.Second):
+			ticker.Stop()
+			if err != nil {
+				return errors.New("Device was not restarted in 60 seconds")
+			}
```
dmissmann, line 153: I think the nicer solution would be to use a `contextContext` that gets passed to this function instead of hardcoding this into the implementation.

Also, this parts is anyways never reached. The ticker fires every 10s, and on the next loop iteration we create a new `time.After` channel (and that's another problem to look out in general 😉 `time.After` always creates a new goroutine, which can't be stopped if we don't need it anymore. Meaning those resources are always only released after the timeout exceeds)

ios/amfi/amfi.go
```diff
@@ -0,0 +1,174 @@
...
+
+	request := map[string]interface{}{"action": 1}
+
+	bytes, err := devModeConn.plistCodec.Encode(request)
+	if err != nil {
+		return fmt.Errorf("EnableDevMode: failed encoding request to service with err: %w", err)
+	}
+
+	err = devModeConn.deviceConn.Send(bytes)
+	if err != nil {
```
dmissmann, line 49: You could check the new `PlistCodecReadWriter` type to simplify that 😉

ios/amfi/amfi.go
shamanec: Actually initially I used 5s, the problem is that the device takes some time to reboot after firing the dev mode request and it was sometimes getting the device and proceeding before it was even rebooted 😓 😅

ios/amfi/amfi.go
shamanec: Yeah, now that I think about it you are actually right, this does reset the timer on each iteration. Using context sounds like a reasonable solution, I might update the code these days. Initially this wrapping function was only in the CLI code and I didn't think much of it when you could just kill it if it hangs, shame on me :D

PR #324, opened 2024-01-11
Allow closing listener in port forwarding

fish-sauce:
### Why
Whenever the `go-ios` is used as a dependency package in any go project, it would be nice to use port forwarding feature with the possibility to free the listening port when the forwarding is stopped

### How
- wrap the `net.Listener` in a dedicated struct
- provide the `Close()` function to free the listening port
- allow the eternal for loop in `connectionAccept` to break when `Close()` is called
- wrap any errors occurred

### DIFF
--- a/ios/forward/forward.go
+++ b/ios/forward/forward.go
@@ -38,6 +38,7 @@ func Forward(device ios.DeviceEntry, hostPort uint16, phonePort uint16) (*ConnLi
 	return cl, nil
 }
 
+// Close stops listening on the host port for the forwarded connection
 func (cl *ConnListener) Close() error {
 	close(cl.quit)
 
--- a/ios/forward/forward.go
+++ b/ios/forward/forward.go
@@ -16,28 +16,54 @@ type iosproxy struct {
 	deviceConn ios.DeviceConnectionInterface
 }
 
+type ConnListener struct {
+	listener net.Listener
+	quit     chan interface{}
+}
+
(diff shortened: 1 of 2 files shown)


ios/forward/forward.go
```diff
@@ -16,28 +16,54 @@ type iosproxy struct {
 	deviceConn ios.DeviceConnectionInterface
 }
 
+type ConnListener struct {
+	listener net.Listener
+	quit     chan bool
```
diegoperini, line 21: This is a single state channel. Bool has 2 states. I recently learned it is possible to define single state channels as

```
quit chan struct{}
```

or

```
testFinishedChannel chan interface{}
```

ios/forward/forward.go
```diff
@@ -16,28 +16,54 @@ type iosproxy struct {
...
+		quit:     make(chan interface{}),
 	}
 
-	go connectionAccept(l, device.DeviceID, phonePort)
+	go connectionAccept(cl, device.DeviceID, phonePort)
+
+	return cl, nil
+}
+
+func (cl *ConnListener) Close() error {
```
dmissmann, line 41: Lets also try to document exported methods 🙂 
```suggestion
// Close stops listening on the host port for the forwarded connection
func (cl *ConnListener) Close() error {
```

dmissmann (approved):

dmissmann (approved):

PR #329, opened 2024-01-30
XCActivityRecord with attachments in xcuitest results

diegoperini:


ios/testmanagerd/testlistener.go
```diff
@@ -90,8 +101,32 @@ func (t *TestListener) testCaseStalled(testClass string, method string, file str
...
+		testCase := t.TestSuite.findTestCase(testClass, testMethod)
+		if testCase == nil {
+			t.TestSuite.TestCases = append(t.TestSuite.TestCases, TestCase{
+				ClassName:  testClass,
+				MethodName: testMethod,
+			})
+			testCase = &t.TestSuite.TestCases[len(t.TestSuite.TestCases)-1]
+		}
+
+		filepath := filepath.Join(os.TempDir(), uuid.New().String())
```
dmissmann, line 114: Lets not use `os.TempDir()` in here. The path where we want to store attachments should be passed to `TestListener` (then we can even handle the case of not storing attachments at all by passing an empty string)

you are using a package name as variable name (you can't use that package after this line in this scope). This should always be avoided (and usually there's a warning in an IDE about this)
```suggestion
		attachmentPath := filepath.Join(os.TempDir(), uuid.New().String())
```

ios/testmanagerd/testlistener.go
```diff
@@ -192,6 +227,27 @@ func (t *TestListener) Done() <-chan struct{} {
...
+}
+
+func (a TestAttachment) Close() error {
+	if _, err := os.Stat(a.Path); errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+
+	return os.Remove(a.Path)
+}
+
```
dmissmann, line 250: And if we inject the path to the `TestListener` we don't need this, as the ownership moves. One can simply delete the directory once they are done
```suggestion
```

dmissmann (changes requested):

ios/testmanagerd/testlistener.go
diegoperini: Removed.

ios/testmanagerd/testlistener.go
diegoperini: Good one, I'll be careful next time. Renamed.

ios/testmanagerd/testlistener.go
```diff
@@ -55,11 +59,20 @@ type TestError struct {
 	Line    uint64
 }
 
-func NewTestListener(logWriter io.Writer, debugLogWriter io.Writer) *TestListener {
+type TestAttachment struct {
```
dmissmann, line 62: the UTI would also be nice to have to get some context of what is handled here

diegoperini: Added

dmissmann (approved):

PR #332, opened 2024-02-02
Ios 17 - support Mac OS only

danielpaulus:
supports fucking iOS17, thanks Apple for making everything harder.


ios/coredevice/coredevice.go
```diff
@@ -0,0 +1,20 @@
+package coredevice
+
+import "github.com/google/uuid"
+
+func BuildRequest(deviceId, feature string, input map[string]interface{}) map[string]interface{} {
```
danielpaulus, line 5: @dmissmann @diegoperini  why is this a one file module?  An explanation would be cool, I can add it to the docs

main.go
```diff
@@ -819,11 +937,21 @@ The commands work as following:
 
 	b, _ = arguments.Bool("reboot")
 	if b {
-		err := diagnostics.Reboot(device)
-		if err != nil {
-			log.Error(err)
+		v, err := ios.GetProductVersion(device)
```
danielpaulus, line 940: @dmissmann @diegoperini  I think I want to move this logic to the module. So others can just call reboot without knowing the device version

PR #333, opened 2024-02-02
add some errors, add generic methods to extract stuff from maps and c…

danielpaulus:
…onvert arrays

### DIFF
--- a/ios/appservice/appservice.go
+++ b/ios/appservice/appservice.go
@@ -17,16 +17,22 @@ import (
 	"howett.net/plist"
 )
 
+// Connection represents a connection to the appservice on an iOS device for iOS17+.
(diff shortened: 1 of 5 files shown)


dmissmann:
> > Fine for me in general. I'm just wondering now if we should aim for exposing only `int` in exported methods to have fewer conversions on that level. We now have `int64` and `uint64` for PIDs (before your changes it was `int64` and `int`, so you didn't make it worse 😄 )
> 
> We can also make it uint64 everywhere which is appropriate for a PID. Negative PIDs are probably never a good idea.

well, [os#Getpid](https://pkg.go.dev/os#Getpid) 😄  or `pid_t` in libc is also int

PR #334, opened 2024-02-02
iOS17 Linux Driver

danielpaulus:
An iOS17 compatible CDC.NCM driver that runs in user space. Use:
 `make build` to build it. 
The resulting ncm-cdc binary needs to run with sudo to work. 
for development use `make up`. It recompiles and runs the binary with sudo so no extra commands needed.

This PR introduces go work spaces to compile a second binary for running golang based network interfaces on linux devices. 

NCM-CDC will detect new ios devices and automatically spin up a new network interface for each device. It assumes libimobiledevice latest usbmuxd is already running.


ncm/usb.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+		if desc.Product < PID_RANGE_LOW || desc.Product > PID_RANGE_MAX {
+			return false
+		}
+		return true
+	})
+	if err != nil {
+		slog.Error("failed opening devices", "err", err)
+	}
+	slog.Info("device list", "length", len(devices))
```
dmissmann, line 72:
```suggestion
	slog.Debug("device list", "length", len(devices))
```

ncm/usb.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+func printStatus() {
+	var connectedDevices []string
+	allocatedDevices.Range(func(key, value interface{}) bool {
+		connectedDevices = append(connectedDevices, key.(string))
+		return true
+	})
+	slices.Sort[[]string](connectedDevices)
+	slog.Info("connected devices", "devices", connectedDevices)
```
dmissmann, line 53:
```suggestion
	slog.Debug("connected devices", "devices", connectedDevices)
```

dmissmann (commented):
works pretty good already 😄

PR #337, opened 2024-02-07
iOS 17 base: Http2, RemoteXpc and device discovery

dmissmann:
Those changes don't add much functionality yet, but they are required later for connecting to RemoteXPC services on iOS 17 devices


danielpaulus (approved):
LGTM

PR #339, opened 2024-02-07
Expose tunnels on HTTP API

dmissmann:


main.go
```diff
@@ -300,14 +301,30 @@ The commands work as following:
...
 	udid, _ := arguments.String("--udid")
 	address, addressErr := arguments.String("--address")
 	rsdPort, rsdErr := arguments.Int("--rsd-port")
 
 	device, err := ios.GetDevice(udid)
-	exitIfError("Device not found: "+udid, err)
-	if addressErr == nil && rsdErr == nil {
-		device = deviceWithRsdProvider(device, udid, address, rsdPort)
+	// device address and rsd port are only available after the tunnel started
+	if !startTunnelCommand {
```
dmissmann, line 316: @danielpaulus this is the part I don't like when it comes to the `start-tunnel` command (and that name can change as well I guess). We try to get a device, while we are simply start a tunnel for all connected devices, and the parameters for `--address` and `--rsd-port` are usable for all other commands, apart from `start-tunnel`

dmissmann (commented):
If that looks fine in general I would start to prepare a PR for tunnels to main

PR #341, opened 2024-02-08
Start QUIC tunnels on iOS 17

dmissmann:
The `tunnel start` command pairs with connected devices (manual pairing) and starts a tunnel for each device. The tunnel parameters (device address and the port for remote service discovery) are exposed via an HTTP-API locally so that other commands can access them without the need to pass those values manually (which is still possible if one wants/needs to use a tunnel created by MacOS)


ios/tunnel/tunnel.go
```diff
@@ -0,0 +1,328 @@
...
+		},
+	)
+	certPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert,
+	})
+	cert5, err := tls.X509KeyPair(certPem, privateKeyPem)
+
+	conf := &tls.Config{
+		InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
dmissmann: That's required from the device side

ios/tunnel/untrusted.go
```diff
@@ -0,0 +1,483 @@
...
+	"crypto/x509"
+	"encoding/base64"
+	"fmt"
+
+	"io"
+
+	"github.com/danielpaulus/go-ios/ios/opack"
+	"github.com/danielpaulus/go-ios/ios/xpc"
+
+	//"github.com/google/uuid"
```
danielpaulus, line 18: this can be removed?

ios/tunnel/untrusted.go
```diff
@@ -0,0 +1,483 @@
...
+		return err
+	}
+
+	ci, err := chacha20poly1305.New(derived)
+	if err != nil {
+		return err
+	}
+
+	signBuf := bytes.NewBuffer(nil)
+	signBuf.Write(key.PublicKey().Bytes())
```
danielpaulus, line 298: you are swallowing a few errors here I think. Since Golang 1.20 it finally supports multierrs out of the box. 
`err := errors.Join(err1, err2)`
You can join them like that if it does not make sense to handle them all individiually.

ios/tunnel/untrusted.go
```diff
@@ -0,0 +1,483 @@
...
+	io.CopyN(buf, hkdfPairSetup, 32)
+	buf.WriteString(t.pairRecords.selfId.Identifier)
+	buf.Write(t.pairRecords.selfId.publicKey())
+
+	signature := ed25519.Sign(t.pairRecords.selfId.privateKey(), buf.Bytes())
+
+	deviceInfo, err := opack.Encode(map[string]interface{}{
+		"accountID":                   t.pairRecords.selfId.Identifier,
+		"altIRK":                      []byte{0x5e, 0xca, 0x81, 0x91, 0x92, 0x02, 0x82, 0x00, 0x11, 0x22, 0x33, 0x44, 0xbb, 0xf2, 0x4a, 0xc8},
+		"btAddr":                      "FF:DD:99:66:BB:AA",
```
danielpaulus, line 423: can you put these things in constants maybe or somehow explain why they are hardcoded?

ios/tunnel/untrusted.go
```diff
@@ -0,0 +1,483 @@
...
+		"mac":                         []byte{0xff, 0x44, 0x88, 0x66, 0x33, 0x99},
+		"model":                       "go-ios",
+		"name":                        "host-name",
+		"remotepairing_serial_number": "remote-serial",
+	})
+
+	deviceInfoTlv := newTlvBuffer()
+	deviceInfoTlv.writeData(typeSignature, signature)
```
danielpaulus, line 431: don'T swallow the errors please

ios/tunnel/tunnel.go
```diff
@@ -0,0 +1,328 @@
+package tunnel
```
danielpaulus, line 1: Please add a detailed description here about what this module is for.

ios/tunnel/codec.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+type pairingData struct {
+	data            []byte
+	kind            string
+	sendingHost     string
+	startNewSession bool
+}
+
+func (p *pairingData) Encode() map[string]interface{} {
```
danielpaulus, line 23: please add a comment explaining how this codec works

ios/tunnel/codec.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func getChildMap(m map[string]interface{}, keys ...string) (map[string]interface{}, error) {
+	if len(keys) == 0 {
+		return m, nil
+	}
+	if c, ok := m[keys[0]].(map[string]interface{}); ok {
+		return getChildMap(c, keys[1:]...)
+	} else {
+		return nil, fmt.Errorf("something went wrong")
```
danielpaulus, line 76: please be more specific with error messages, add the func name f.ex.

ios/tunnel/codec.go
```diff
@@ -0,0 +1,218 @@
...
+	return nil
+}
+
+func getChildMap(m map[string]interface{}, keys ...string) (map[string]interface{}, error) {
+	if len(keys) == 0 {
+		return m, nil
+	}
+	if c, ok := m[keys[0]].(map[string]interface{}); ok {
+		return getChildMap(c, keys[1:]...)
+	} else {
```
danielpaulus, line 75: I think this else is not needed

ios/tunnel/codec.go
```diff
@@ -0,0 +1,218 @@
...
+			"_0": map[string]interface{}{
+				"request": map[string]interface{}{
+					"_0": req,
+				},
+			},
+		},
+	})
+}
+
+func (c *controlChannelReadWriter) write(message map[string]interface{}) error {
```
danielpaulus, line 138: would be nice to explain a bit what these functions do and what the difference between write and write Request ist

ios/opack/main.m
```diff
@@ -0,0 +1,70 @@
+//go:build ignore
+
+/*
+xcrun clang -DDECODE -fobjc-arc -fmodules -F /System/Library/PrivateFrameworks/ -framework CoreUtils main.m --output decode
```
danielpaulus, line 4: what is this for?

ios/tunnel/tunnel_api.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return fmt.Errorf("UpdateTunnels: failed to get list of devices: %w", err)
+	}
+	for _, d := range devices.DeviceList {
+		udid := d.Properties.SerialNumber
+		if _, exists := m.tunnels[udid]; exists {
+			continue
+		}
+		log.WithField("udid", udid).Info("start tunnel")
+		t, err := m.ts.StartTunnel(context.TODO(), d, m.pm)
```
danielpaulus, line 119: 4yi:
lol, I was thinking "what is context.TODO(), and what is the difference between this and context.Background()??"
turns out, there is none:
```var (

	background = new(emptyCtx)

	todo       = new(emptyCtx)

)


// Background returns a non-nil, empty Context. It is never canceled, has no

// values, and has no deadline. It is typically used by the main function,

// initialization, and tests, and as the top-level Context for incoming

// requests.

func Background() Context {

	return background

}


// TODO returns a non-nil, empty Context. Code should use context.TODO when

// it's unclear which Context to use or it is not yet available (because the

// surrounding function has not yet been extended to accept a Context

// parameter).

func TODO() Context {

	return todo

}
```

danielpaulus (changes requested):
This is really damn excellent work! 
I just have a couple of nitpicks, and would like some more comments. 
If something exposes a REST API, which I think makes sense in this scenario, there should be a nice CLI command to consume the REST API as well. I don't want to force users to be messing around with cURL unless they want to.

ios/tunnel/untrusted.go
dmissmann: `Write` on a `bytes.Buffer` never returns an error. I'm pretty sure it's handled in other places, but it's actually not necessary

ios/tunnel/untrusted.go
dmissmann: As mentioned in the other comment, there will never be an error as `tlvBuffer` uses a `bytes.Buffer` internally

dmissmann:
> This is really damn excellent work!
I just have a couple of nitpicks, and would like some more comments.
If something exposes a REST API, which I think makes sense in this scenario, there should be a nice CLI command to consume the REST API as well. I don't want to force users to be messing around with cURL unless they want to.

Thanks! Users shouldn't be required to interact with the API in general, and on the go-ios side it's already implemented that this is not needed. We try to attach this info to the device https://github.com/danielpaulus/go-ios/blob/70d5f9a6b4084d6d7d1f00a38a0f1606c7776e97/main.go#L321

main.go
```diff
@@ -223,6 +233,10 @@ The commands work as following:
    ios timeformat (24h | 12h | toggle | get) [--force] [options] Sets, or returns the state of the "time format". iOS 11+ only (Use --force to try on older versions).
    ios diskspace [options]											  Prints disk space info.
    ios batterycheck [options]                                         Prints battery info.
+   ios start-tunnel [options] [--pair-record-path=<pairrecordpath>]   Creates a tunnel connection to the device. If the device was not paired with the host yet, device pairing will also be executed.
```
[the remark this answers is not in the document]
dmissmann: It's not `tunnel start` and there's also `tunnel ls` to list the running tunnels with their address and RSD port, not sure if there's anything else we can provide

ios/tunnel/tunnel.go
dmissmann: There's now a `doc.go` file for the package documentation with some detail. It's a first draft, and I'll probably update it again 😄

ios/tunnel/tunnel_api.go
dmissmann: removed that part 🙂 we take the context as an argument now. However, we also create a new child context based on the one we get with 10s for tunnel creation. Should we also expose this as an argument?

ios/tunnel/untrusted.go
danielpaulus: ```Write appends the contents of p to the buffer, growing the buffer as needed. The return value n is the length of p; err is always nil. If the buffer becomes too large, Write will panic with [ErrTooLarge](https://pkg.go.dev/bytes#ErrTooLarge).```
oh man.. eyebleed. But ok, then. I did not know this. Maybe nice to add a comment to explain this. It is quite unusualy to not handle errors.

ios/tunnel/doc.go
```diff
@@ -0,0 +1,47 @@
+// Package tunnel implements creating a TUN interface for iOS 17 devices. These tunnels are used to connect to services
```
danielpaulus, line 1: I think the convention is to put this docs into [packagename].go. So tunnel.go in this case

ios/tunnel/tunnel.go
danielpaulus: ah, I think the golang convention is to put this here. I might be also wrong tho, let me know

ios/tunnel/doc.go
dmissmann: both ways work. However, if a package has a longer text (which is of course subjective) it can/should go into the `doc.go` ([godoc](https://go.dev/blog/godoc)). And from my point of view if you open the file and you have to scroll way down to get to the actual code, it's long enough to go into it's own file 🙂

danielpaulus (approved):
Amazing!

danielpaulus (approved):

corrrso:
Amazing work! Out of curiosity: When is this going to be released to npm?

PR #343, opened 2024-02-09
Update and reorder prepare skip keys

shamanec:
* Updated the skip keys for fresh device preparation according to the official Apple [docs](https://developer.apple.com/documentation/devicemanagement/skipkeys)


danielpaulus (approved):
Nice one!

PR #345, opened 2024-02-14
add basic metrics for Prometheus

danielpaulus:
This PR adds basic metrics, exposed with Prometheus.
go-ncm accepts one cmd line argument now: --prometheusport=8080 
- if specified, prometheus metrics will be available at http://0.0.0.0:prometheusport/metrics
- if not specified, the prometheus endpoint will not be started and not be available.

### Metrics
1. device_count
Type: Gauge
Description: Tracks the number of iOS devices currently connected to the system. A gauge is a metric that represents a single numerical value that can arbitrarily go up and down.
Usage: Helps in monitoring the fluctuation in the number of connected iOS devices over time.
2. network_receive_bytes
Type: CounterVec
Description: A counter metric that represents the cumulative bytes received on the virtual TAP device. It is labeled by device and serial to provide granular tracking across different devices.
Labels:
device: network interface name
serial: The serial number of the iOS device.
Usage: Useful for analyzing the volume of incoming network traffic on a per-device basis.
3. network_send_bytes
Type: CounterVec
Description: A counter metric for tracking the total bytes sent to the virtual TAP device. Similar to network_receive_bytes, it uses device and serial labels for detailed monitoring.
Labels:
device: network interface name
serial: The serial number of the iOS device.
Usage: Allows for the assessment of outgoing network traffic per device, aiding in network performance evaluation.
4. usb_send_bytes
Type: CounterVec
Description: Represents the total bytes sent through the USB endpoint to connected devices. This metric is labeled by serial to differentiate between devices.
Labels:
serial: The serial number of the iOS device.
Usage: Facilitates tracking of data sent to devices over USB, which is crucial for understanding USB throughput and identifying potential bottlenecks.
5. usb_receive_bytes
Type: CounterVec
Description: Counts the bytes received from iOS devices over the USB endpoint. It is labeled with serial, allowing for device-specific traffic analysis.
Labels:
serial: The serial number of the iOS device.
Usage: Essential for monitoring the amount of data received from devices over USB, offering insights into USB data receipt efficiency and potential issues.


ncm/usb.go
```diff
@@ -29,6 +29,7 @@ var deviceLock = sync.Mutex{}
 var deviceCounter = 0
 
 func Start(c chan os.Signal) error {
+	go startPrometheus(13333)
```
diegoperini, line 32: It would be nice to have this configurable via env or CLI args. Not crucial. 👍

danielpaulus: I think that makes sense. I'll add that.

PR #346, opened 2024-02-15
Added appservice package

diegoperini:
`appservice` package contains necessary code to launch apps on iOS 17.


danielpaulus (approved):
nice! Thanks @diegoperini

PR #353, opened 2024-02-20
Add debug proxy for tunnel interface

dmissmann:
We can simply use a [pcap]() library to capture the communication going over the tunnel interface (it't not encrypted at that point).

usbmuxd and utun captured traffic are put into separate subdirectories in a dump. Their content is also slightly different. With utun we already put the service name in the directory name, and we try to decode the content when possible (currently we do this for `http2`, `RemoteXPC` and `DTX`)


ios/debugproxy/utun/decoding.go
```diff
@@ -0,0 +1,138 @@
...
+		if m.IsFileOpen() {
+			log.Info("file transfer started, skipping remaining data ")
+			return nil
+		}
+	}
+}
+
+func decodeRemoteDtx(w io.Writer, r io.Reader) error {
+	for {
+		m, err := dtx.ReadMessage(r)
```
diegoperini, line 125: I wonder if we can make this work with fragmented messages out of box, using `DecodeNonBlocking()` just like it's used by `TestFragmentedMessage` unit tests.

ios/debugproxy/utun/decoding.go
```diff
@@ -0,0 +1,138 @@
...
+		return unknown
+	}
+	if string(b) == "PRI " {
+		return http2
+	}
+	i := binary.LittleEndian.Uint32(b)
+	if i == 0x29b00b92 {
+		return remoteXpc
+	}
+	if string(b[:3]) == "y[=" {
```
diegoperini, line 48: Isn't this magic also 4 bytes long?

ios/debugproxy/utun/connection.go
```diff
@@ -0,0 +1,157 @@
...
+			inErr := decodeHttp2FrameHeaders(incoming.w, inFile, false)
+			return errors.Join(outErr, inErr)
+		})
+		_, _ = outFile.Seek(0, io.SeekStart)
+		_, _ = inFile.Seek(0, io.SeekStart)
+		return createDecodingFiles(dir, "http.bin", func(outgoing, incoming pair) error {
+			outErr := decodeHttp2(outgoing.w, outFile, true)
+			inErr := decodeHttp2(incoming.w, inFile, false)
+			if err := errors.Join(outErr, inErr); err != nil {
+				//return err
```
diegoperini, line 111: Do we want to keep this?

ios/debugproxy/utun/capture_noop.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/danielpaulus/go-ios/ios"
+)
+
+func Live(ctx context.Context, iface string, provider ios.RsdPortProvider, dumpDir string) error {
+	return errors.New("capturing traffic on the utun interface is only supported on MacOS")
```
diegoperini, line 13: Does this still hold true?

diegoperini: Can go-ios sniff tunnels created by itself running on a separate process?

dmissmann: it should work on linux as well. however, after changing it to build everywhere the build fails on linux. And I think initially the motivation was that this anyways is mainly used on MacOS I would expect that you always want Xcode/MacOS to perform a certain action that we want to capture

danielpaulus:
Thank you for this, @dmissmann 🙏 — this is genuinely valuable work. A darwin-side sniffer that reassembles the iOS 17 tunnel and auto-decodes HTTP/2, RemoteXPC, and DTX is exactly the kind of debugging capability go-ios has been missing, and there's nothing else like it.

I'm closing the PR only because the branch has been overtaken by later refactors and can't be adopted as-is:
- The `dproxy` command was moved into the command registry (`runDproxyCommand` in `cmd_device_debug.go`), and main already advertises the `--mode`/`--iface`/`--address`/`--rsd-port` flags — so this is a partial rewrite rather than a rebase.
- The new `utun/` files use logrus, which the tree removed in the `golog` migration (won't compile against main).
- A couple of real bugs to carry over when re-landing: `ifaceAddr` indexes `Addresses[1]` (panics on single-address interfaces), there's no TCP reassembly yet (reorders/retransmits corrupt the decode), `xpc.jsonl` isn't newline-delimited, and the capture dirs/files are `0777` while holding decrypted traffic (want `0700`/`0600`).

The decoders themselves are the valuable part and are very close. I'd love to see a fresh PR that re-scaffolds just `ios/debugproxy/utun/` against current main — wired into the existing `runDproxyCommand`/flags, on `golog`, with gopacket reassembly + a BPF filter and unit tests over canned packet bytes. I'm very happy to help with the plumbing so you can focus on the decoders. Thanks again — this deserves to land in some form. 💙

PR #354, opened 2024-02-20
Added device info service and CLI

diegoperini:


main.go
```diff
@@ -128,6 +129,7 @@ Usage:
   ios batterycheck [options]
   ios tunnel start [options] [--pair-record-path=<pairrecordpath>]
   ios tunnel ls [options]
+  ios deviceinfo [options] (display | lockdown)
```
danielpaulus, line 132: @diegoperini does it make sense to merge this with the ios info command?

diegoperini: It does but it's also not a straightforward merge since the output is entirely different and we don't know who we are gonna make unhappy if we don't keep the old one. What's more is deviceinfo is a real service, named after the one listening on the device and it has/will have more domains (like display, lockdown) in the future. The merge will be a never ending problem as we introduce new domains. I'd be in favor of freezing the old one in time and deprecating it.

@dmissmann Any thoughts?

diegoperini: Removed "deviceinfo" command and merged the behavior into "info". 

Usage:
```shell
# Prints same output as before, fetching the data from lockdown
ios info

# Same as above but more verbose invocation
ios info lockdown

# Newly added, fetches the data from RSD services over the tunnel
# Address and rsd port are mandatory, prints a useful error message when omitted
ios info display --address=<tunnelAddress> --rsd-port=<rsdPort>
```

main.go
```diff
@@ -557,7 +558,27 @@ The commands work as following:
 
 	b, _ = arguments.Bool("info")
 	if b {
-		printDeviceInfo(device)
+		if display, _ := arguments.Bool("display"); display {
+			if address == "" || rsdPort == 0 {
+				log.Fatal("Please provide --address and --rsd-port to retreive display information")
+				return
+			}
+
```
dmissmann, line 566:
```suggestion
```
This needs to be removed, those arguments are not required, and if go-ios was used to start a tunnel it's handled implicitly

dmissmann (changes requested):

main.go
```diff
@@ -557,7 +558,27 @@ The commands work as following:
...
+			deviceInfo, err := deviceinfo.NewDeviceInfo(device)
+			exitIfError("Can't connect to deviceinfo service", err)
+			defer deviceInfo.Close()
+
+			info, err := deviceInfo.GetDisplayInfo()
+			exitIfError("Can't fetch dispaly info", err)
+
+			log.WithField("display", info).Info("Got display info")
```
dmissmann, line 574:
```suggestion
			fmt.Println(convertToJSONString(info))
```
to have the same output format

dmissmann (changes requested):

main.go
diegoperini: Removing that check makes errors impossible to read because it looks like a tunnel failure with an internal stacktrace printed to the output. Are you sure?

diegoperini: This is the error when those checks are removed:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13e8f92]

goroutine 1 [running]:
github.com/danielpaulus/go-ios/ios.ConnectToXpcServiceTunnelIface({0x15, {0xc000693748, 0x8}, {0x1c9c3800, {0xc000693810, 0x3}, 0x15, 0x14200000, 0x12a8, {0xc0000aa1e0, ...}}, ...}, ...)
        /Users/diegoperini/Projects0x77BC187E/go-ios/ios/connect.go:126 +0x32
github.com/danielpaulus/go-ios/ios/deviceinfo.NewDeviceInfo({0x15, {0xc000693748, 0x8}, {0x1c9c3800, {0xc000693810, 0x3}, 0x15, 0x14200000, 0x12a8, {0xc0000aa1e0, ...}}, ...})
        /Users/diegoperini/Projects0x77BC187E/go-ios/ios/deviceinfo/deviceinfo.go:17 +0x65
main.Main()
        /Users/diegoperini/Projects0x77BC187E/go-ios/main.go:567 +0x1da5
main.main()
        /Users/diegoperini/Projects0x77BC187E/go-ios/main.go:61 +0xf
exit status 2
```

How can we make it more human friendly?

diegoperini: Option 1: I can just document it in the help message, explaining tunnels are mandatory for display info.

diegoperini: Option 2: We add additional validation pass before all commands that cannot work without tunnels. i.e `if command not in TUNNEL_REQUIRING_COMMANDS_SLICE then quitEarlyWithNiceErrors()`

dmissmann: We need to check for `nil` where we get the panic and to make it more readable for the user is to check if the rsdProvider is set on the device or not and print a message on how to fix this (and the preferred way there is to start the tunnel with `go-ios tunnel start`)

diegoperini: Done.

danielpaulus: nice, I really like the UX of this

danielpaulus (approved):

dmissmann (approved):

PR #356, opened 2024-02-21
Use the shim service over the tunnel interface for zipconduit and syslog

dmissmann:
Services that still work over usbmuxd on iOS 17 are also exposed as 'shim' services via RSD over the tunnel interface. With this change we transparently switch to using those services if available.


danielpaulus (approved):
Nice!

PR #361, opened 2024-02-26
Report proxy dispatcher panics under test listener

diegoperini:
**Why?**
DTX message dispatcher has a lot of assumptions about received messages, all hard-coded. Unexpected payloads panic during decoding, which creates more problems for the users of the programmatic API.

**What is affected?**
No change in the runtime behavior of the happy path of message dispatcher. Unhappy paths now has errors.

**How was this fixed?**
We now utilize the `err` property under `TestListener` to report all decoder and message dispatcher errors in the test results.

### DIFF
--- a/ios/nskeyedarchiver/unarchiver.go
+++ b/ios/nskeyedarchiver/unarchiver.go
@@ -2,6 +2,7 @@ package nskeyedarchiver
 
 import (
 	"fmt"
+	"runtime/debug"
 
 	log "github.com/sirupsen/logrus"
 	plist "howett.net/plist"
@@ -14,7 +15,8 @@ import (
 func Unarchive(xml []byte) (result []interface{}, err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			err = fmt.Errorf("Unarchive: %s", r)
+			stacktrace := string(debug.Stack())
+			err = fmt.Errorf("Unarchive: %s\n%s", r, stacktrace)
 		}
 	}()
 
--- a/ios/nskeyedarchiver/unarchiver.go
+++ b/ios/nskeyedarchiver/unarchiver.go
@@ -11,7 +11,13 @@ import (
 // Primitives will be extracted just like regular Plist primitives (string, float64, int64, []uint8 etc.).
 // NSArray, NSMutableArray, NSSet and NSMutableSet will transformed into []interface{}
 // NSDictionary and NSMutableDictionary will be transformed into map[string] interface{}. I might add non string keys later.
-func Unarchive(xml []byte) ([]interface{}, error) {
+func Unarchive(xml []byte) (result []interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("Unarchive: %s", r)
+		}
+	}()
+
 	SetupDecoders()
 	plist, err := plistFromBytes(xml)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ios/testmanagerd/proxydispatcher.go
```diff
@@ -17,6 +18,13 @@ type proxyDispatcher struct {
 }
 
 func (p proxyDispatcher) Dispatch(m dtx.Message) {
+	defer func() {
+		if r := recover(); r != nil {
+			p.testListener.err = fmt.Errorf("Dispatch: %s", r)
+			p.testListener.TestRunnerKilled()
```
mk-ahmed-saucelabs, line 24: We don't know whether the TestRunner has been killed, only a message dispatch failed. Unless part of the recovery process is to kill the TestRunner.

dmissmann: Yes, this should definitely not introduce the side effect of killing the runner.

And are you trying to recover from a potential panic in the `extract...` functions of this file? If that's the case please use safe [Type Assertions](https://go.dev/ref/spec#Type_assertions) (that should be the default anyways I guess)

ios/nskeyedarchiver/unarchiver.go
```diff
@@ -11,7 +11,13 @@ import (
 // Primitives will be extracted just like regular Plist primitives (string, float64, int64, []uint8 etc.).
 // NSArray, NSMutableArray, NSSet and NSMutableSet will transformed into []interface{}
 // NSDictionary and NSMutableDictionary will be transformed into map[string] interface{}. I might add non string keys later.
-func Unarchive(xml []byte) ([]interface{}, error) {
+func Unarchive(xml []byte) (result []interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("Unarchive: %s", r)
```
dmissmann, line 17: you can make the error a bit more helpful with adding the stack to the error (https://pkg.go.dev/runtime/debug#Stack)

ios/testmanagerd/proxydispatcher.go
diegoperini: > We don't know whether the TestRunner has been killed

That's why we assume it's killed or should be killed by us. There is no partial recovery that can save a test run from a panic.

ios/nskeyedarchiver/unarchiver.go
diegoperini: Thank you, added.

ios/testmanagerd/proxydispatcher.go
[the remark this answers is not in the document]
diegoperini: THANK YOU. Luckily, all properties of `p` we mutated except `err` were reference types, preventing this bug causing unexpected outcomes. Pushed a fix.

diegoperini: > Yes, this should definitely not introduce the side effect of killing the runner.

Removed it.

> And are you trying to recover from a potential panic in the extract... functions of this file? 

Yes but not due to type assertions. The panic recovered in `Unarchive()` can still cause nil derefence errors after the fact.

dmissmann (approved):

ios/testmanagerd/proxydispatcher.go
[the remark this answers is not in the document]
diegoperini: > Or recover here will proceed the code execution again?
It will but we will have missing data in the test report. It's best effort.

PR #363, opened 2024-02-26
Added support for fragmented DTX messages

diegoperini:
**Why?**
We never had to deal with fragmented DTX messages before. That's okay for messages of short length, which are pretty much all messages that do not contain large pieces of binary data. However this is not enough if we are gonna implement messages that nest data of arbitrary sizes, i.e file transfers.

**What is affected?**
Our DTX decoder is affected. Unlike before, It'll no longer ignore fragments when they arrive over the tunnel. It use to only consider the first fragment of a family of fragmented messages and ignore the rest. 

**How was this fixed?**
When reading bytes from the serial connection, we now check the DTX message header to see how many fragments are gonna be sent by the device. If the number of fragments is greated than 1, we pause decoding the initial message and iteratively collect each fragment, appending their payload to the original (paused) message. Once all fragments are received, we resume decoding the original message as if it's received within a single fragment.


ios/dtx_codec/decoder.go
```diff
@@ -24,22 +24,34 @@ func ReadMessage(reader io.Reader) (Message, error) {
...
+			return Message{}, NewOutOfSync(fmt.Sprintf("Wrong Fragment Index: %d/%d", result.FragmentIndex, result.Fragments))
 		}
-		// 32 offset is correct, the binary starts with a payload header
-		messageBytes := make([]byte, result.MessageLength)
-		_, err := io.ReadFull(reader, messageBytes)
-		if err != nil {
-			return Message{}, err
+
+		// Read all fragments and merge payloads under result.fragmentBytes
+		for i := 1; i < int(result.Fragments); i++ {
```
dmissmann, line 34: Maybe @danielpaulus can answer this. Can there be other messages between fragments? Because with this implementation we expect that to be the case, and before that there was a separate fragment decoder (which would now be dead code I assume https://github.com/danielpaulus/go-ios/blob/main/ios/dtx_codec/channel.go#L148)

diegoperini: > Can there be other messages between fragments?
I don't know but I had the same worry. Tried to create that situation by running parallel tests first time I created this solution. I haven't succeeded in producing such behavior. All fragments arrive sequentially, with no messages injected in between as far as I can tell.

diegoperini: My understanding is Apple also relies on this serial behavior since multiple concurrent connections are already multiplexed at the lower IP layer, over different sockets.

diegoperini: Updated the branch, utilizing existing Daniel's defragmenter already present in channel.go.

danielpaulus (approved):
Maybe consider making the message type uint32? seems good to go tho

diegoperini:
> Maybe consider making the message type uint32? seems good to go tho

Done.

PR #364, opened 2024-02-26
Support for NSData, NSMutableData test attachments with graceful ignore for unknown types

diegoperini:
Builds one top of https://github.com/danielpaulus/go-ios/pull/363 since attachments are almost always received via fragmented DTX messages.


ios/dtx_codec/encoder.go
```diff
@@ -74,7 +74,7 @@ func writeHeader(messageBytes []byte, messageLength uint32, Identifier int, Conv
 	binary.LittleEndian.PutUint32(messageBytes[28:], expectsReplyUint32)
 }
 
-func writePayloadHeader(messageBytes []byte, MessageType int, payloadLength int, auxLength int) {
+func writePayloadHeader(messageBytes []byte, MessageType MessageType, payloadLength int, auxLength int) {
 	binary.LittleEndian.PutUint32(messageBytes, uint32(MessageType))
```
danielpaulus, line 78: @diegoperini  why don't you make it an alias for uint32 and stop the casting?

danielpaulus (approved):
Maybe consider changing the MessageType to uint32. looks good otherwise

ios/dtx_codec/encoder.go
diegoperini: We define a function on top of `MessageType` which maps numbers to human friendly names under the `String()` function. I'm kinda new to golang but I believe we cannot do that on type aliases. See here: https://github.com/danielpaulus/go-ios/pull/364/files#diff-e74aeb8496b7de791c62636d22da05ac8451409e098d70365e3c3d84a1cf48afR76

diegoperini:
> Maybe consider changing the MessageType to uint32. looks good otherwise

Done

PR #367, opened 2024-02-29
Deleting system attachments after running tests, preventing garbage piling up

diegoperini:


dmissmann (approved):

dmissmann:
putting those numbers into constants would probably worth doing to explain what 0,1,2 represent 😉

diegoperini:
> putting those numbers into constants would probably worth doing to explain what 0,1,2 represent 😉

Good idea. Added those named constants.

PR #370, opened 2024-03-05
increase the number of random bytes sent before the parameter exchange

dmissmann:
the lower value caused the device to close the tunnel on iOS 17.4 beta


diegoperini (approved):
How the hell did you figure this out? 👏 👏 👏 Approved.

PR #371, opened 2024-03-05
Added proper support for tests without apps under test

diegoperini:
Apps that don't have a corresponding app under test, such as WDA require an unnecessary bundle id argument provided in the CLI and programmatic API.

This PR removes that requirement by making `--bundle-id` optional. When omitted, test config excludes necessary keys related to target app, so that such tests can run simply with the following command:

```shell
ios runtest --test-runner-bundle-id=com.facebook.WebDriverAgentRunner.xctrunner --xctest-config=WebDriverAgentRunner.xctest"
```


ios/testmanagerd/xcuitestrunner.go
```diff
@@ -236,7 +236,7 @@ func RunXCUITest(bundleID string, testRunnerBundleID string, xctestConfigName st
 	}
 	info, err := getAppInfos(bundleID, testRunnerBundleID, apps)
```
dmissmann, line 237: lets rather have `getAppInfos` do it a bit different. At the moment we are getting the infos for both the app under test and the test runner at the same time. Lets split this and get only one of them at a time. The we can simply not call it when not needed, and we are still able to handle errors if there's something wrong with the provided bundle id

ios/nskeyedarchiver/objectivec_classes.go
```diff
@@ -108,6 +108,15 @@ func NewXCTestConfiguration(
 		testIdentifiersToSkipEntry = createTestIdentifierSet(productModuleName, testsToSkip)
 	}
 
+	var appUnderTestExists = targetApplicationPath != "" && targetApplicationBundleID != ""
+	if appUnderTestExists {
+		contents["productModuleName"] = productModuleName
+		// contents["targetApplicationArguments"] = []interface{}{} //TODO: triggers a bug
+		contents["targetApplicationBundleID"] = targetApplicationBundleID
+		// contents["targetApplicationEnvironment"] = //TODO: triggers a bug map[string]interface{}{}
```
dmissmann, line 116: Which bugs? That would be worth noting down 😉

diegoperini: Those are really old lines from years ago. Removed to avoid confusion.

ios/testmanagerd/xcuitestrunner.go
diegoperini: Good idea. Pushed an update, please review.

ios/testmanagerd/xcuitestrunner.go
```diff
@@ -234,9 +234,9 @@ func RunXCUITest(bundleID string, testRunnerBundleID string, xctestConfigName st
 	if err != nil {
 		return make([]TestSuite, 0), fmt.Errorf("RunXCUITest: cannot browse user apps: %w", err)
 	}
-	info, err := getAppInfos(bundleID, testRunnerBundleID, apps)
+	info, err := getTargetAppInfo(bundleID, apps)
 	if err != nil {
-		return make([]TestSuite, 0), fmt.Errorf("RunXCUITest: cannot get app information: %w", err)
+		bundleID = ""
```
dmissmann, line 239: shouldn't this be the case only if `bundleID` was already empty? If we provide a bundle id and this fails that's an error, isn't it?

ios/testmanagerd/xcuitestrunner.go
```diff
@@ -504,42 +530,60 @@ func createTestConfigOnDevice(testSessionID uuid.UUID, info testInfo, houseArres
...
+
+type targetAppInfo struct {
 	targetAppPath       string
 	targetAppBundleName string
 	targetAppBundleID   string
 }
 
-func getAppInfos(bundleID string, testRunnerBundleID string, apps []installationproxy.AppInfo) (testInfo, error) {
-	info := testInfo{}
+func getTestAppInfo(testRunnerBundleID string, apps []installationproxy.AppInfo) (testAppInfo, error) {
```
dmissmann, line 558: why two different functions and two different structs for getting app infos?

you could collect all infos in a single struct, and thus with a single function
```
type appInfo struct {
	path       string
	bundleName string
	bundleID   string
	homePath   string
}

func getappInfo(bundleID string, apps []installationproxy.AppInfo) (appInfo, error) {
	for _, app := range apps {
		if app.CFBundleIdentifier == bundleID {
			info := appInfo{
				path:       app.Path,
				bundleName: app.CFBundleName,
				bundleID:   app.CFBundleIdentifier,
			}
			if home, ok := app.EnvironmentVariables["HOME"].(string); ok {
				info.homePath = home
			}
			return info, nil
		}
	}

	return appInfo{}, fmt.Errorf("Did not find test app for '%s' on device. Is it installed?", testRunnerBundleID)
}
```

ios/testmanagerd/xcuitestrunner.go
diegoperini: True. Altered the `if{}` block to only fetch data when needed.

ios/testmanagerd/xcuitestrunner.go
diegoperini: Thank you 🙇 Applied.

dmissmann (approved):

PR #385, opened 2024-03-14
feat: add font size

DavidPerezP124:
This adds the ability to update the font size when using ax, and in the future the rest of the settings can be added.

It currently can only be set once, but in the future could be updated during the goroutine's lifetime.

If we end the accessibility the font size will revert to the default.

Setting logic from:
https://github.com/doronz88/pymobiledevice3/blob/0e67465fe7757748777c497170bdae84502a8d8c/pymobiledevice3/services/accessibilityaudit.py#L318


danielpaulus (approved):
Nice work, thanks @DavidPerezP124 !

PR #398, opened 2024-04-12
Get all XCUITest logs

dmissmann:
Starting with iOS 17 XCUITest logs are sent over two channels. There are debug logs that come over the DTX connection (like they did in the past), but the non-debug messages are sent over a different connection to the `com.apple.coredevice.openstdiosocket` service.

We now open a connection to the openstdiosocket-service before we launch the test runner, and we tell the appservice to which socket it should connect stdio. And after that we pipe the data we get from the stdio-socket to the logoutput.


ios/testmanagerd/xcuitestrunner.go
```diff
@@ -358,18 +359,26 @@ func runXUITestWithBundleIdsXcode15Ctx(
...
 
-	pid, err := startTestRunner17(device, appserviceConn, "", testRunnerBundleID, strings.ToUpper(testSessionID.String()), info.testApp.path+"/PlugIns/"+xctestConfigFileName, args, env)
+	testRunnerLaunch, err := startTestRunner17(device, appserviceConn, "", testRunnerBundleID, strings.ToUpper(testSessionID.String()), info.testApp.path+"/PlugIns/"+xctestConfigFileName, args, env)
 	if err != nil {
 		return make([]TestSuite, 0), fmt.Errorf("runXUITestWithBundleIdsXcode15Ctx: cannot start test runner: %w", err)
 	}
 
+	defer testRunnerLaunch.Close()
+	go func() {
+		_, err := io.Copy(testListener.logWriter, testRunnerLaunch)
```
diegoperini, line 369: Does this imply an infinite loop? How is it going to behave when `Read()` on `testRunnerLaunch` fails with an error different than `io.EOF`?

ios/appservice/appservice.go
```diff
@@ -48,6 +50,27 @@ type AppLaunch struct {
 	Pid int
 }
 
+// AppWithStdIo is the launched app with a connection to the stdio-socket
+type AppWithStdIo struct {
+	stdIoConnection openstdio.Connection
+	AppLaunch
```
diegoperini, line 56: Now that we this, are there any remaining reasons to have support for app launch without stdio? Maybe that would help us only have one type to represent launches instead of this embedding.

ios/testmanagerd/xcuitestrunner.go
dmissmann: It copies all there is available until an error happens, and it returns the error if it's not `io.EOF` ([docs](https://pkg.go.dev/io#Copy))

for us that means we get a log if a real error happens. I don't think there's much more we can do here as a new connection to the stdio service will have a new UDID and we can't change that on the existing app. And killing the whole test run because of this sound also a bit harsh (it could be a config option for the test run, but I don't want to add another argument to the methods here for that)

ios/appservice/appservice.go
dmissmann: I'll give the way it's represented here another look, but the reason why I didn't simply replace it was that if you start the app with stdio you also have to close this connection when you're done. And that means "fire-and-forget" app launches wouldn't be so nice

diegoperini (approved):

PR #417, opened 2024-06-24
Use a userspace TUN device to enable running commands without sudo

danielpaulus:
This PR removes the need for starting a tunnel with sudo using gVisor's user space networking stack. 
`ios tunnel start --userspace` will start a userspace only networking stack that will work without sudo and you can use devices. Pairing does not yet work here, will look into it next. 
`sudo ios tunnel start`  will use normal OS TUN devices for this as before.

There is now an experimental mode to automatically start go-ios tunnel management. If you set the `ENABLE_GO_IOS_AGENT` env var to `yes` it will automatically start the tunnel manager agent and you can run commands without any prep.


shamanec (approved):

ios/tunnel/tunnel_api.go
```diff
@@ -17,8 +20,53 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-// DefaultHttpApiPort is the port on which we start the HTTP-Server for exposing started tunnels
-const DefaultHttpApiPort = 28100
+var netClient = &http.Client{
+	Timeout: time.Millisecond * 200,
+}
+
+func IsAgentRunning() bool {
```
dmissmann, line 27:
```suggestion
func isAgentRunning() bool {
```

ios/tunnel/tunnel_api.go
```diff
@@ -17,8 +20,53 @@ import (
...
+	resp, err := netClient.Get(fmt.Sprintf("http://%s:%d/health", "127.0.0.1", ios.DefaultHttpApiPort()))
+	if err != nil {
+		return false
+	}
+	return resp.StatusCode == http.StatusOK
+}
+func WaitUntilAgentReady() bool {
```
dmissmann, line 34:
```suggestion
func waitUntilAgentReady() bool {
```

ios/connect.go
```diff
@@ -333,3 +362,39 @@ func initializeXpcConnection(h *http.HttpConnection) error {
...
+}
+
+// defaultHttpApiPort is the port on which we start the HTTP-Server for exposing started tunnels
+// 60-105 is leetspeek for go-ios :-D
+const defaultHttpApiPort = 60105
+
+// DefaultHttpApiPort is the port on which we start the HTTP-Server for exposing started tunnels
+// if GO_IOS_AGENT_PORT is set, we use that port. Otherwise we use the default port 60106.
+// 60-105 is leetspeek for go-ios :-D
+func DefaultHttpApiPort() int {
```
dmissmann, line 386:
```suggestion
func HttpApiPort() int {
```
it's only the default if no environment variable is set

ios/connect.go
```diff
@@ -333,3 +362,39 @@ func initializeXpcConnection(h *http.HttpConnection) error {
...
+		return defaultHttpApiPort
+	}
+	return port
+}
+
+// DefaultTunnelPort is the port on which you can connect to services on the device.
+// It is used for connecting to the user space TUN device.
+// It will be on DefaultHttpApiPort() + 1
+// Which means 60106 per default
+func DefaultTunnelPort() int {
```
dmissmann, line 398: if `GO_IOS_AGENT_PORT` can be set, this one should be configurable too

ios/connect.go
```diff
@@ -199,18 +220,26 @@ func ConnectToTunnel(device DeviceEntry, port int) (*net.TCPConn, error) {
 	return conn, nil
 }
 
-func ConnectToHttp2WithAddr(a string, port int) (*http.HttpConnection, error) {
-	addr, err := net.ResolveTCPAddr("tcp6", fmt.Sprintf("[%s]:%d", a, port))
-	if err != nil {
-		return nil, fmt.Errorf("ConnectToHttp2WithAddr: failed to resolve address: %w", err)
-	}
+func ConnectToHttp2WithAddr(a string, port int, userSpaceTUN bool) (*http.HttpConnection, error) {
```
dmissmann, line 223: having the argument `userSpaceTUN` all over the place doesn't feel nice. We should try to either move this into an interface, or have separate functions with proper names for the two ways of connecting to the device

ios/tunnel/rwcendpoint.go
```diff
@@ -0,0 +1,158 @@
+package tunnel
+
+// I took this code from https://github.com/xjasonlyu/tun2socks which is an amazing project
+// but GPL licensed.
+// Technically this would require me from having go-ios be MIT licensed now.
```
dmissmann, line 5: go-ios has the MIT license 😉 It would need to be GPL licensed, and that would be a problem.

ios/tunnel/rwcendpoint.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// once is used to perform the init action once when attaching.
+	once sync.Once
+
+	// wg keeps track of running goroutines.
+	wg sync.WaitGroup
+}
+
+// New returns stack.LinkEndpoint(.*Endpoint) and error.
+func RWCEndpointNew(rw io.ReadWriter, mtu uint32, offset int) (*Endpoint, error) {
```
dmissmann, line 50: Is this the endpoint that basically exposes the userspace tunnel?

ios/tunnel/tunnel_api.go
```diff
@@ -241,8 +315,13 @@ type deviceLister interface {
 type manualPairingTunnelStart struct {
 }
 
-func (m manualPairingTunnelStart) StartTunnel(ctx context.Context, device ios.DeviceEntry, p PairRecordManager, version *semver.Version) (Tunnel, error) {
+func (m manualPairingTunnelStart) StartTunnel(ctx context.Context, device ios.DeviceEntry, p PairRecordManager, version *semver.Version, userspaceTUN bool) (Tunnel, error) {
 	if version.Major() >= 17 && version.Minor() >= 4 {
+		if userspaceTUN {
```
dmissmann, line 320: that would be one place where it's easy to use separate interfaces for the two tunnel types. The `tunnelStarter` was only made for unit testing (and I just realized that they were just removed), but now we could have two different implementations for the tunnel types

ios/tunnel/rwcendpoint.go
danielpaulus: yeah it was late when I wrote it, thanks for the comment!

ios/tunnel/rwcendpoint.go
danielpaulus: No this is used to connect the virtual TUN device to the lockdown connection readwritecloser

danielpaulus: I added a comment to clarify.

ios/tunnel/tunnel_api.go
danielpaulus: Yeah I am not a big fan of these Java/Mockito style whitebox unit tests that just ensure the right methods were called. Imo, if you have to use a mock, it's probably not worth testing it.

ios/connect.go
danielpaulus: This one actually needs to change so every device gets its own port. Right now, all conns go to the same device

ios/connect.go
dmissmann: It's probably a bit to much for this PR, but how about adding methods for opening up connections to the device to `DeviceEntry`. It's always a device that we want to connect to and a `DeviceEntry` basically represents that

ios/connect.go
dmissmann: Ah ok, at the moment we can have one device per agent process? I assume that will change in the future and this here will become a port range?

ios/connect.go
danielpaulus: It is something I actually tried yesterday. Either this or a connection module taking a device as a param. I like the DevicrEntry being a simple entity without too much stuff attached

dmissmann: That's also fine, assuming that you mean package and not module

danielpaulus: Yes.

PR #419, opened 2024-06-27
send one packet at a time for the lockdown tunnel

dmissmann:
the TUN interface often returned an error when an incomplete IPv6 packet was sent (one IPv6 packet + parts of the next one) sending one packet at a time resolved this issue

### DIFF
--- a/ios/tunnel/tunnel_lockdown.go
+++ b/ios/tunnel/tunnel_lockdown.go
@@ -1,7 +1,9 @@
 package tunnel
 
 import (
+	"bufio"
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


ios/tunnel/tunnel_lockdown.go
```diff
@@ -87,21 +89,38 @@ func forwardTUNToDevice(ctx context.Context, mtu uint64, tun io.Reader, deviceCo
...
 		case <-ctx.Done():
 			return nil
 		default:
-			n, err := deviceConn.Read(b)
+			_, err := io.ReadFull(br, ip6Header)
+			if err != nil {
+				return fmt.Errorf("failed to read IPv6 header: %w", err)
+			}
+
+			if ip6Header[0] != 0x60 {
```
danielpaulus, line 109: @dmissmann  I think IPv4 can happen here, no big deal and should not need an error

dmissmann: Currently that shouldn't be possible, we configure the TUN interface only with an IPv6 https://github.com/danielpaulus/go-ios/blob/main/ios/tunnel/tunnel.go#L184

danielpaulus: hmm ok, but I definitely got some ipv4 packets on mac os when testing this.

danielpaulus: I might add code to just drop ipv4 then later.

danielpaulus (approved):
nice!

PR #422, opened 2024-07-08
ensure that the test listener channel gets closed only once

dmissmann:
the close call happens in multiple places and we can not be sure that only one of those methods is getting called during an execution


danielpaulus (approved):
Thanks @dmissmann!

PR #424, opened 2024-07-11
fix memory leaks

dmissmann:
There was a memory leak when we handle fragmented DTX messages. The fragments were not completely cleaned up.

And in XCUITests the attachments were also retained completely, even though we were only using the string values of those attachments. Cloning those strings resolves this


diegoperini (approved):
Wow, nasty! Thanks for fixing this.

PR #425, opened 2024-07-11
Remove screenshotr and fix screenshot service

shamanec:


ios/instruments/screenshot.go
```diff
@@ -62,12 +95,12 @@ func startConversionQueue() {
...
 }
 
-func startScreenshotting(conn *Connection) {
+func startScreenshotting(conn *ScreenshotService) {
 	for {
 		start := time.Now()
 		pngBytes, err := conn.TakeScreenshot()
 		if err != nil {
-			log.Fatal("screenshotr failed", err)
+			log.Fatal("Screenshot failed", err)
```
danielpaulus, line 103: can you find a way to remove logFatal? Don't know why I added this, but the library should not log.Fatal

danielpaulus (approved):
Nice @shamanec !

PR #435, opened 2024-07-19
Draft: Added a buffered reader for dtx connections

diegoperini:


ios/deviceconnection.go
```diff
@@ -31,11 +34,12 @@ type DeviceConnectionInterface interface {
 type DeviceConnection struct {
 	c               net.Conn
 	unencryptedConn net.Conn
+	bufferedReader  io.Reader
```
dmissmann, line 37: I don't think that's the right place to add a buffered reader. We have two connections to the device already in this struct. If we add buffering now to one of them, the buffer may hold more data than expected and switching between those sessions could break (I think this is mainly happening on older iOS versions)

If you want to add buffering to the DTX connection as the title says, it would be safer to do so in the DTX connection itself as nothing about the connection would change at that point anymore.

diegoperini: I think I understand. 

That would mean I may have to define a new type that implements `net.Conn`, wraps a real `net.Conn` and uses a buffered reader internally for the `Read()` operations. The rest of the operations would be delegated to the wrapped, original net.Conn instance. This should happen BEFORE we construct a `DeviceConnection` object. 

Correct?

diegoperini:
Closing this one. Alternative solution here: https://github.com/danielpaulus/go-ios/pull/438/files

PR #438, opened 2024-07-23
Added a buffered reader to dtx connections

diegoperini:


ios/dtx_codec/decoder_test.go
```diff
@@ -230,6 +231,31 @@ func TestDecoder(t *testing.T) {
 		assert.Equal(t, uint32(425), msg.PayloadHeader.AuxiliaryLength)
 		assert.Equal(t, uint32(596), msg.PayloadHeader.TotalPayloadLength)
 		assert.Equal(t, uint32(0), msg.PayloadHeader.Flags)
+	}
+}
+
+func TestDecoderWithBufferedReader(t *testing.T) {
```
dmissmann, line 237: Isn't that identical to the test `TestDecoder` above? since this one has less duplication I would remove the one above. And each test that calls `ReadMessage` uses a buffered reader now, not only this test 😉 that part can be removed from the name

diegoperini: > Isn't that identical to the test TestDecoder above? 

It was identical with the exception of unbuffered reader in the original test. 

I applied your suggestions. 👍

dmissmann (approved):

PR #439, opened 2024-07-23
Automatically connect to os TUN device without constructing a device entry

diegoperini:


danielpaulus:
@diegoperini can you explain why this is needed please?

diegoperini:
> @diegoperini can you explain why this is needed please?

[This commit](https://github.com/danielpaulus/go-ios/commit/8c6f14e2675011d5a8e71cf0b7e4c92a60c57bda) removed the programmatic API for connecting to a tunnel "without" having to construct a device entry object. 

This PR recovers the ability to run the tunnel in one process and connect to it in another in a way that lets the second process skip reading device info from usbmuxd and RSD entirely. When the second process can interface with the device just with its IP address and the port, it also allows moving the second process into another computer (or container, vm etc) as long as they are in the same network.

danielpaulus (approved):

danielpaulus:
@diegoperini  merging the other PRs created a small conflict, can you please resolve it?

PR #440, opened 2024-07-30
move initializing the buffered reader out of the loop

dmissmann:
having it inside the loop destroys the reader after every iteration and it may contain bytes at that point


shamanec (approved):

danielpaulus (approved):
Nice!!

PR #445, opened 2024-08-03
fix: launch app with options

sam80180:


danielpaulus (approved):
Nice addition! Thanks!

PR #451, opened 2024-08-08
Add userspace port CLI option

shamanec:
* When trying to run a command while manually supplying `rsd-port` and `address` and using userspace tunnels, also the userspace tunnel port has to be supplied or it doesn't work

### DIFF
--- a/main.go
+++ b/main.go
@@ -149,6 +149,7 @@ Options:
   >                         This parameter is similar to '--address' and can be obtained by the same log filter
   --proxyurl=<url>          Set this if you want go-ios to use a http proxy for outgoing requests, like for downloading images or contacting Apple during device activation.
   >                         A simple format like: "http://PROXY_LOGIN:PROXY_PASS@proxyIp:proxyPort" works. Otherwise use the HTTP_PROXY system env var.
+  --userspace-port=<port>   Optional. Set this if you run a command supplying rsd-port and address and your device is using userspace tunnel
 
 The commands work as following:
 	The default output of all commands is JSON. Should you prefer human readable outout, specify the --nojson option with your command.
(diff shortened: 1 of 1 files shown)


danielpaulus:
@dmissmann  can you take a look?

dmissmann (commented):
Not sure if I understand why there can be a problem. When `rsd-port` and `address` are provided we are pointing to a tunnel that is not a userspace tunnel (for example I usually use it when I want to connect to the tunnel that is opened by MacOS). Why should there be a userspace tunnel at the same time?

shamanec:
> Not sure if I understand why there can be a problem. When `rsd-port` and `address` are provided we are pointing to a tunnel that is not a userspace tunnel (for example I usually use it when I want to connect to the tunnel that is opened by MacOS). Why should there be a userspace tunnel at the same time?

My use case was that I needed finer control over tunnels so I was starting a tunnel per device using the module but then I wanted to run a command using the CLI providing `rsd-port` and `address` - but it doesn't work if the tunnel is a userspace tunnel. In the end I resolved it by using the modules everywhere around my app instead of using the CLI so I don't much care now if this gets merged. It might be good if someone wants finer control or is testing something out

dmissmann:
> > Not sure if I understand why there can be a problem. When `rsd-port` and `address` are provided we are pointing to a tunnel that is not a userspace tunnel (for example I usually use it when I want to connect to the tunnel that is opened by MacOS). Why should there be a userspace tunnel at the same time?
> 
> My use case was that I needed finer control over tunnels so I was starting a tunnel per device using the module but then I wanted to run a command using the CLI providing `rsd-port` and `address` - but it doesn't work if the tunnel is a userspace tunnel. In the end I resolved it by using the modules everywhere around my app instead of using the CLI so I don't much care now if this gets merged. It might be good if someone wants finer control or is testing something out

ah ok, got it. and since it's an optional parameter it's perfectly fine 👍

dmissmann (approved):

danielpaulus:
LGTM, Thanks Nikola :-D

PR #399, opened 2024-04-12
feat: Polish and fix the NPM package

mykola-mokhnach:
These are preparation steps needed to include this tool into appium-xcuitest-driver dist.

The list of changes made in this PR:
- A "fat" binary is now created for macOS, which supports both arm64 and x64 architectures
- A proper symlink to the go-ios binary is created upon package install. index.js acts as a proxy for the symlink, whose main purpose is to match (and cache) the current platform needed to run the binary (actually the target binary path) and to act as an execution forwarder.
- Added README, so we could properly reflect the package info at https://www.npmjs.com/package/go-ios
- Limited the package to Node 18/ Npm 8 and later versions

PR #452, opened 2024-08-09
Support installation for arm64 architecture

jdragonbae:
This patch supports installation for arm64 architecture.
Since go-ios supports arm64 in #399, 
we can use arm64 instead of amd64 for installation.

### DIFF
--- a/npm_publish/postinstall.js
+++ b/npm_publish/postinstall.js
@@ -12,7 +12,8 @@ var path = require('path'),
 var ARCH_MAPPING = {
     "ia32": "386",
     "x64": "amd64",
-    "arm": "arm"
+    "arm": "arm",
+    "arm64": "arm64"
 };
 
 // Mapping between Node's `process.platform` to Golang's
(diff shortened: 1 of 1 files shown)


danielpaulus (approved):
Thanks!

jdragonbae:
@danielpaulus Can we merge this PR into main?

danielpaulus:
@jdragonbae  I had to revert #399 because it broke the release pipeline. I will try and double check this week that the change above works with the reverted state too.

danielpaulus:
Thank you for this, @jdragonbae — and apologies it sat so long. 🙏 You were absolutely right that Apple Silicon shouldn't be stuck on an amd64/Rosetta binary, and this PR pointed at the real gap.

Closing it now because the underlying issue has since been solved a different way and the change is effectively a no-op against current `main`:

- `ARCH_MAPPING` already includes `"arm64": "arm64"` on `main`.
- The release workflow now builds a **universal (`lipo`) macOS binary** and ships that *same* binary to **both** the `darwin-amd64` and `darwin-arm64` dist folders (release.yml: `lipo ... -create -output bin/ios`, then `cp ./mac-bin/ios` into both). So even the legacy "using amd64 build on M1 mac" fallback path now installs a binary that runs **natively** on Apple Silicon — no Rosetta.

The only remnant is that the `"using amd64 build on M1 mac"` log line is now misleading (it's a universal binary, not amd64), which we'll tidy up separately.

So the practical outcome you wanted — native arm64 installs on Apple Silicon — already ships. Thanks again for flagging it and for your patience! 🎉

PR #455, opened 2024-08-10
Avoid panic due to sending on closed channel

nicolasbouffard:
I came across panics due to sends that might still happen after the channel has been close here. After investigation, it seems to me that it is perfectly safe to leave this channel oppened as, once it's no longer used, the garbage collector should clean it, as per [this SO](https://stackoverflow.com/a/8593986).

Now that I think of it, I wonder if it's to pass a channel with no capacity to the `checkEntry` goroutine since, should `FindDeviceInterfaceAddress` return, there would be nobody left to read what `checkEntry` could be trying to send, potentially leading to a leaking goroutine. Since there is no wait on it, the program still runs though.

### DIFF
--- a/ios/discover.go
+++ b/ios/discover.go
@@ -20,7 +20,10 @@ func FindDeviceInterfaceAddress(ctx context.Context, device DeviceEntry) (string
 	}
 
 	result := make(chan string)
-	defer close(result)
+
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	defer cancel()
 
 	for _, iface := range ifaces {
 		resolver, err := zeroconf.NewResolver(zeroconf.SelectIfaces([]net.Interface{iface}), zeroconf.SelectIPTraffic(zeroconf.IPv6))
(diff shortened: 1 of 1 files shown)


dmissmann:
> ow that I think of it, I wonder if it's to pass a channel with no capacity to the checkEntry goroutine since, should FindDeviceInterfaceAddress return, there would be nobody left to read what checkEntry could be trying to send, potentially leading to a leaking goroutine. Since there is no wait on it, the program still runs though.

Yep, I also think that would happen. And then we have a goroutine leak **and** the channel is also not garbage collected 😄 

Nevertheless, you're right and the `defer close(result)` should not be there. The go doc is also pretty clear about that ` It should be executed only by the sender, never the receiver`. `FindDeviceInterfaceAddress` is the receiver. So if we want to close it `checkEntry` should do that, but only doing that would block the goroutine as the value is never read.

What we could do to resolve it is
1. `FindDeviceInterfaceAddress` creates a child context with `context.WithCancel(ctx)` and this child is cancelled when `FindDeviceInterfaceAddress` returns (and this child context is passed to `checkEntry`)
2. `checkEntry` passes the conext to `tryHandshake`
3. `tryHandshake` only writes to the `result` channel if the context has not been cancelled yet

dmissmann (approved):
> it would require synchronization between all senders in order to close it once all senders returned, which seems a bit overkill to me if we keep in mind that not closing channels should be fine thanks to the garbage collector.

yeah that would just introduce unnecessary complexity for no gain as the garbage collector takes care of it.

thanks for fixing this!

PR #458, opened 2024-08-10
Fix panic index out of range

nicolasbouffard:
This PR fixes panics I came across due to an ignored error and and index out of range error.


danielpaulus (approved):
Really nice fix!

PR #459, opened 2024-08-10
Allow listing rsd services

nicolasbouffard:
This PR adds methods and commands for listing RSD services from a device.


danielpaulus (approved):
nice addition!

danielpaulus (approved):

Issue #460, opened 2024-08-12
write unix /var/run/usbmuxd: write: broken pipe


dokisha:
**Describe the bug**

This bug has been happening for years now, but it happens so random that it was hard to deduce what is the root cause.

When go-ios is used normally it takes some time for the bug to occur, but when it does the device is permanently corrupted and needs to be rebooted.

Commands like `ps`, `kill`, and `runwda` stop working (but probably every command that uses lockdown)

**To Reproduce**

1. Open **two** terminals

2. Copy paste this code in **both** terminals and run it simultaneously
```
for n in {1..200}
do
    if go-ios ps --apps --udid=SOME_UDID 2>&1>error.log; then
        echo "Success"
    else
        echo "Command failed"
        cat error.log
    fi
done
```
3. Wait a bit
4. Lockdown/communication gets broken
5. Commands (ps and runwda) start getting response:
```
{"level":"debug","msg":"Failed connecting to com.apple.instruments.remoteserver, trying com.apple.instruments.remoteserver.DVTSecureSocketProxy"}
{"EnableServiceSSL":true,"Port":60328,"Request":"StartService","Service":"com.apple.instruments.remoteserver.DVTSecureSocketProxy","level":"debug","msg":"Service started on device"}
{"channel_id":"com.apple.instruments.server.services.deviceinfo","level":"debug","msg":"Requesting channel"}
{"level":"debug","msg":"DTX Connection with EOF"}
{"level":"debug","msg":"i0.0 c0 t:Ack mlen:0 aux_len0 paylen0"}
{"channel_id":"com.apple.instruments.server.services.deviceinfo","error":"Timed out waiting for response for message:5 channel:0","level":"error","msg":"failed requesting channel"}
{"channel_id":"com.apple.instruments.server.services.deviceinfo","level":"debug","msg":"Channel open"}
{"level":"error","msg":"DeviceConnection failed writing 204 bytes, only 0 sent"}
{"level":"error","msg":"Failed sending: write unix @-\u003e/var/run/usbmuxd: write: broken pipe"}
{"channel_id":"com.apple.instruments.server.services.deviceinfo","error":"write unix @-\u003e/var/run/usbmuxd: write: broken pipe","level":"info","methodselector":"runningProcesses","msg":"failed starting invoking method"}
panic: runtime error: index out of range [0] with length 0

```
I really hope you can reproduce it and figure out what is causing this.

**Expected behavior**
Lockdown doesn't get broken and device doesn't need restart.

[...]

PR #462, opened 2024-08-13
Avoid concurrent requests for getting RSD service for a device

nicolasbouffard:
This PR prevents concurrent requests for getting RSD service for a device, based on the trio formed by address, port, and TUN port (even if it's zero). I did this because I noticed my program getting stuck multiple times in this function and ended up writing a debug program that calls the function multiple times asynchronously and got a freeze every single time, whereas running it synchronously or by preventing concurrent access gave 100% of success rate.

This might be related to #460 in the sense that this is a problem that may never happen if we are "lucky" and no concurrent accesses are made, but that probability increases dramatically when more requests are made in parallel. The issue is most likely not related to go-iOS but it might be the easiest place to "fix" it, at the cost of loosing some parallelization capability.

### DIFF
--- a/ios/rsd.go
+++ b/ios/rsd.go
@@ -6,11 +6,15 @@ import (
 	"io"
 	"net"
 	"strconv"
+	"sync"
 
 	"github.com/danielpaulus/go-ios/ios/http"
 	"github.com/danielpaulus/go-ios/ios/xpc"
 	log "github.com/sirupsen/logrus"
 )
+
+// _requestsMap stores a mutex for every request attempt to the trio formed by address, port, and TUN port. This allows to make sure that no more than one request is made at a time to a given trio, since doing so can lead to stuck requests and, therefore, stuck programs and/or goroutine leaks.
+var _requestsMap = sync.Map{}
 
 // RsdPortProvider is an interface to get a port for a service, or a service for a port from the Remote Service Discovery on the device.
 // Used in iOS17+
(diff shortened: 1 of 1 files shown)


dmissmann:
I would rather move the responsibility of locking something to a place where no global state needs to be added. The only place I can think of where the concurrent access is happening is when we start a tunnel (`getUntrustedTunnelServicePort`) at the same time as connecting to a device for which we go a MDNS response (`FindDeviceInterfaceAddress`).
Did you see the issues when you were running `go-ios tunnel start`?

danielpaulus:
Thank you for this, @nicolasbouffard 🙏 — and especially for the standalone repro repo. A reliably-reproducible concurrency hang is gold, and you correctly identified that firing multiple RSD requests at one device wedges the process.

I'm closing the patch itself because the serialization lands a layer too shallow to be the right fix:
- The mutex is released when the constructor returns, but the blocking read that actually hangs is in `Handshake()`, which runs *after* — so the lock stops just short of the call most likely to block.
- Keyed on `(addr,port[,tunport])` rather than the device, the two entry points (`NewWithAddrPort` vs `NewWithAddrPortDevice`) can key the *same* connection differently and not serialize across each other.
- And holding a lock across the network handshake turns 'one stuck request' into 'every request to that device wedged', plus the global map never evicts.

The root cause is that the RSD dial + HTTP/2 + handshake reads have **no deadline** anywhere (`ios/connect.go`, `ios/http/http.go`, `ios/rsd.go`) — so a device that accepts the socket but never replies hangs forever, and `FindDeviceInterfaceAddress`'s fan-out even ignores `ctx`. The fix that actually cures your hang is a dial timeout + read deadline / ctx plumbing on that path (a stuck peer then errors out fast). If you're up for redirecting the PR that way — and turning your repro into a regression test — I'd be thrilled to review it. If you'd rather I take it from here, I'll credit you. Thanks again for the sharp diagnosis and the repro. 💙

PR #464, opened 2024-08-15
Add Support for XCTest

bahrimootaz:
- Provide the ability to run an XCTEST by passing the flag `--xctest` to `ios runtest` 
- Test the ability to run a XCUITEST on a Real Device
- Test the ability to run a XCTEST on a Real Device

by the default, if the flag `--xctest` is not provided we will run XCUITEST. so to be able to run an XCTEST we need the flag to be present.


si-net:
I think the big question for me is if we can detect that a file is a XCTest or a XCUiTest and automatically set this flag, or if we should leave this decision to clients. 

We tried to detect this by looking at the file extension, but this does not seem to work.

danielpaulus:
@bahrimootaz this looks really nice, merging the other PRs has created a small conflict. Can you please resolve it?

dmissmann:
> I think the big question for me is if we can detect that a file is a XCTest or a XCUiTest and automatically set this flag, or if we should leave this decision to clients.
>
> We tried to detect this by looking at the file extension, but this does not seem to work.

it can be detected when you have access to the `.ipa` file. The `Info.plist` of the `.xctest` bundle contains a entry `"XCTContainsUITests" => 1` for XCUITests. But in the `testmanagerd` package we work with apps that are already installed and we can't access this file anymore (or at least I don't know of a way to do that)

bahrimootaz:
> @bahrimootaz this looks really nice, merging the other PRs has created a small conflict. Can you please resolve it?

I rebased master and resolved the conflict. Can you please take a look again.

bahrimootaz:
> > I think the big question for me is if we can detect that a file is a XCTest or a XCUiTest and automatically set this flag, or if we should leave this decision to clients.
> > We tried to detect this by looking at the file extension, but this does not seem to work.
> 
> it can be detected when you have access to the `.ipa` file. The `Info.plist` of the `.xctest` bundle contains a entry `"XCTContainsUITests" => 1` for XCUITests. But in the `testmanagerd` package we work with apps that are already installed and we can't access this file anymore (or at least I don't know of a way to do that)

After Checking a lot of Docs I think it's safe to say: Accessing the `Info.plist` file of an already installed app is restricted due to the security and sandboxing policies of the operating system.

danielpaulus (approved):
LGTM

PR #488, opened 2024-09-29
fix(Makefile): use native GOOS/GOARCH by default

ardnew:
The Makefile attempts to determine the target OS/arch so that it can specify `GOOS`/`GOARCH` explicitly when calling `go`.

This doesn't really achieve anything as it ends up competing with the same default behavior of `go` as well as the user's own configured environment.

Furthermore, [it causes the build to fail](https://github.com/user-attachments/files/17181175/go-ios.make.txt) on newer Apple silicon Macs, where `uname -m` reports `arm64` (not `aarch64` expected by the Makefile).

This change removes the behavior that always-overrides `GOARCH`. The user can either specify `GOOS` and `GOARCH` in the Makefile where commented, or they can simply export the variables through their environment and the Makefile will respect them.

Otherwise, do not override `GOARCH` or `GOOS` within the Makefile. Let the `go` toolchain figure it out as intended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,30 +9,19 @@
 GO_IOS_BINARY_NAME=ios
 NCM_BINARY_NAME=go-ncm
 
+# Define only if compiling for system different than our own
+OS=
+ARCH=
 
-# Detect the system architecture
-UNAME_S := $(shell uname -s)
-UNAME_M := $(shell uname -m)
-
-# Default GOARCH value
-GOARCH := amd64
-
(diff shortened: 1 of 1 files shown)


aluedeke:
@danielpaulus that one would be nice to merge

danielpaulus:
Sorry folks for the long time to merge, I was out for health reasons for quite a bit. Back now.

danielpaulus (approved):
Great fix, thanks!

PR #501, opened 2024-10-16
parse syslog

aluedeke:
parse all fields from the syslog into dedicated json fields when passing `--parse` so filtering gets easier
stick to the legacy formats if nothing is passed or json disabled
the timestamp handling is really ugly as syslog doesn't provide the year 🤯 nor the timezone

### DIFF
--- a/main.go
+++ b/main.go
@@ -915,8 +915,7 @@ The commands work as following:
 		}
 
 		rawTestlog, rawTestlogErr := arguments.String("--log-output")
-		env := arguments["--env"].([]string)
-
+		env := splitKeyValuePairs(arguments["--env"].([]string), "=")
 		isXCTest, _ := arguments.Bool("--xctest")
 
 		if rawTestlogErr == nil {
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -2064,36 +2067,113 @@ func printDeviceInfo(device ios.DeviceEntry) {
...
+
+			fmt.Println(logFormatter(logMessage))
 		}
 	}()
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 	<-c
 }
 
+func rawSyslog(log string) string {
```
danielpaulus, line 2104: can you abye move this logic out of the main and to the syslog package? 
If someone wants to use this from the API later or just as a module having it in main.go is kind of an antipattern.
And yes, I know I am guilty of this myself :-D

aluedeke: i would not move the raw and legacy parser anywhere, raw doesn't do anything and the legacy json parser is pretty much useless ;) so offering the proper parser for reuse makes sense everything else should be hidden, could even be marked as deprecated. Let me know if that makes sense

danielpaulus: oh yeah. Now I see it. Thanks for explaining!

danielpaulus (approved):

Issue #502, opened 2024-10-22
Deadlock in userspace tunnel due to the connection closing on one part.


stoktamisoglu:
**Describe the bug**
When there is high throughput like running **a loop to launch/kill app** and at the same time getting syslog with another **ios syslog** process then there is a high chance that the userspace tunnel will be deadlocked or get into circular waiting lock situation.
 
**To Reproduce**
1. Start a userspace tunnel with **ios tunnel start --userspace** within a console/terminal
2. Run a script like the below on Windows Powershell (a Linux script can be also used to reproduce)
```
for ($i = 1; $i -le 500; $i++) {

    Write-Host "Iteration ${i}: Launching"
    C:\dev\goios146\ios.exe launch com.apple.mobilesafari -v -t # com.apple.Maps
    
    Start-Sleep -Seconds 0.5

    Write-Host "Iteration ${i}: Killing"
    C:\dev\goios146\ios.exe kill com.apple.mobilesafari -v -t
    Start-Sleep -Seconds 0.5
}
```
3. When the script runs, open a new console/terminal and run

`ios syslog > syslog.txt`

The script running in the loop hangs and the tunnel also stalls or totally hangs.

As it is a race condition issue, you may need to stop syslog with Ctrl+C and start again perhaps a few times. 

There are some other ways to reproduce which can be provided upon a request.

**Expected behavior**
Tunnel should continue working and the script loop along with syslog shouldn't be stalled.

**Desktop (please complete the following information):**
 - OS: Windows, Linux

**Smartphone (please complete the following information):**
 - Device: iPhone XR
 - OS: iOS18.1

**Additional context**
I narrowed down the problem to the functions :

```
func proxyConns(rw1 io.ReadWriter, rw2 io.ReadWriter) error {
	err1 := make(chan error)
	err2 := make(chan error)
	go ioCopyWithErr(rw1, rw2, err1)
	go ioCopyWithErr(rw2, rw1, err2)
	return errors.Join(<-err1, <-err2)
}

func ioCopyWithErr(w io.Writer, r io.Reader, errCh chan error) {
	_, err := io.Copy(w, r)
	errCh <- err
}
```
With some debugging I observe that the connection is getting closed after the one of the io.Close in ioCopyWithErr then the second io.Copy is not aware the one part of the connection is closed and waiting on its 	
`_, err := io.Copy(w, r)
`
Because it doesn't reach to send to **unbuffered channel**	
`errCh <- err
`
[...]

PR #503, opened 2024-10-22
Improved userspace tunnel concurrency to avoid race conditions.

stoktamisoglu:
The problem is described in the issue #502. A potential solution to this problem is to close the reader and writer after the io,Copy() and sending result (EOF, error) to the error channel. By closing the writer and reader the counter part io.Copy() will drain and unblocked. By this change the code will be resistant to one side network connection closes.

The close function needs to be synced with sync.Once to avoid multiple entrance so double freeing.

Apart from this it is essential to use buffered channel like 

`	errCh := make(chan error, 2)
`

Because, buffered channel sender and receiver should be synced at the same time. This means if the code doesn't reach

 	return errors.Join(<-err1, <-err2)

at the same time with the line in ioCopyWithErr

`	errCh <- err
`

Then it will be dead/circular locked.

I've tested it with different scenarios as mentioned in #502 and it worked very performant and with no locking issue at all.

I'm not a golang expert but after some extensive research this solution seems robust.

### DIFF
--- a/ios/diagnostics/diagnostics.go
+++ b/ios/diagnostics/diagnostics.go
@@ -33,6 +33,27 @@ func Reboot(device ios.DeviceEntry) error {
 	return service.Close()
 }
 
+// Battery extracts the battery ioregistry stats like Temperature, Voltage, CurrentCapacity
+func (diagnosticsConn *Connection) Battery() (IORegistry, error) {
+	req := newIORegistryRequest()
+	req.addClass("IOPMPowerSource")
+
+	reader := diagnosticsConn.deviceConn.Reader()
+	encoded, err := req.encoded()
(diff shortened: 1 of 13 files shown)


stoktamisoglu:
Hi @danielpaulus @dmissmann @shamanec , would you please review the proposed fix as it fixed several problems when the tunnel is overwhelming particularly in a loop to launch ios with launchapp, killapp, ps  and apps.

danielpaulus (approved):
@stoktamisoglu this is an amazing fix! Thanks for the great writeup and PR description.

stoktamisoglu:
Thanks a lot @danielpaulus for prompt reviewing and new release.

PR #507, opened 2024-10-29
Put XCTest parameters into a config struct

dmissmann:
This is a follow up of https://github.com/danielpaulus/go-ios/pull/506
I noticed there that we already quite a few arguments that we pass around in the XCTest methods. To clean that up I put them into a struct. (also, with https://github.com/danielpaulus/go-ios/pull/464 support for unit tests was added and the functions were still called XCUITest)


danielpaulus (approved):
Nice refactor!

danielpaulus:
@dmissmann  can you resolve the conflicts?

PR #510, opened 2024-11-08
Fix stuck XCUITest runner when a popup is shown on the screen

dmissmann:
When a popup (like a permission popup for the camera, notifications, etc) is shown before the XCUITest is started, the XCUITest runner will hang and after 30s we get the error `Failed to background test runner within 30.0s.`

By adding the option `"__ActivateSuspended": uint64(1),` when we launch the XCUITest (like Xcode does), this is not happening anymore.

Also, the `Error()` method of `NSError` was a bit too verbose. In the error that we got here, there is a entry `screenshot-data` in the `userInfo` dictionary, and we shouldn't include that in the error message (it's still stored in the error if someone really needs that)

### DIFF
--- a/ios/nskeyedarchiver/objectivec_classes.go
+++ b/ios/nskeyedarchiver/objectivec_classes.go
@@ -586,7 +586,11 @@ func NewNSError(object map[string]interface{}, objects []interface{}) interface{
 }
 
 func (err NSError) Error() string {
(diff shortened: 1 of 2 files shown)


shamanec (approved):
Amazing find, I spent some effort on this but never would have thought of that prefix... 🚀

PR #515, opened 2024-11-18
Check for non-empty 'Error' string in 'StartSessionResponse'

buxomant:
Issue: When sending a `StartSession` request, we just check if there was an error reading the response message. However, we don't check if the response object itself contains an `Error` field.

Fix: Start checking for the `Error` field, and if present return an error accordingly.

### DIFF
--- a/ios/startsession.go
+++ b/ios/startsession.go
@@ -2,6 +2,7 @@ package ios
 
 import (
 	"bytes"
+	"errors"
 
 	plist "howett.net/plist"
 )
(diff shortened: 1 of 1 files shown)


ios/startsession.go
```diff
@@ -52,6 +54,9 @@ func (lockDownConn *LockDownConnection) StartSession(pairRecord PairRecord) (Sta
 		return StartSessionResponse{}, err
 	}
 	response := startSessionResponsefromBytes(resp)
+	if response.Error != "" {
+		return StartSessionResponse{}, errors.New(response.Error)
```
dmissmann, line 58: lets add a bit more context to the error
```suggestion
		return StartSessionResponse{}, fmt.Errorf("failed to start new lockdown session: %s", response.Error)
```

dmissmann (approved):

PR #518, opened 2024-11-25
Allow ResponseWithReturnValueInPayload messages to be passed to dtxConnection.Dispatch

cziter15:
Allow ResponseWithReturnValueInPayload messages to be passed to dtxConnection.Dispatch.
That makes them to be able to reach the Message Dispatcher and this is useful in some cases (instruments).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/cziter15/go-ios
+module github.com/danielpaulus/go-ios
 
 go 1.22.0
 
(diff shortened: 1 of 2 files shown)


sarxos:
Hi @danielpaulus Can we merge this? Without this change the startSamplingAtTimeInterval does not work well and we do not receive FPS metrics because they are ignored and dispatcher is not invoked.

PR #526, opened 2024-12-06
Removed error log for iOS <17 trying to start tunnel

joao-borges:
**Description**

For an environment with mixed older iOS devices (v < 17) and newer ones (v >= 17) there is a persistent non stop log in the tunnel process that says tunnel was not created for iOS older version. This is a redundant information but to not lose the logging statement, I've moved it to a trace level and made the necessary changes to support that.


aluedeke:
@joao-borges can you split the PR in 2 please? i like both changes (there are a lot of things we should fix in docs) but handling them separately makes things just easier.

joao-borges:
@aluedeke thanks for the feedback, I modified this PR to handle only the older iOS logs, and I opened a new one for the documentation changes: https://github.com/danielpaulus/go-ios/pull/535

joao-borges:
@danielpaulus what would be the path to get this merged?

aluedeke (approved):

danielpaulus:
Thank you for this, @joao-borges 🙏 — the annoyance you were fixing was real: in a mixed fleet, iOS <17 devices spamming a tunnel-start error every poll cycle is genuinely noisy, and you were right that it needed handling.

I'm closing it because the problem has since been solved a different way, and this diff would now work against that solution:
- #682 added a `failedDevices` exponential backoff (30s→5min) to `UpdateTunnels`, so an unsupported device now fails once and is throttled instead of logged every cycle — the acute spam is gone.
- That backoff keys on the tunnel-start returning an **error**. This PR changes the unsupported-iOS path to return `(Tunnel{}, nil)` — a success — which would bypass the backoff and re-probe (and re-open a usbmux socket) every ~1s, reintroducing the exact socket leak #682 closed.
- The branch also predates the logrus→`golog` migration, so it no longer applies cleanly.

If you'd still like the log fully silent for iOS <17 (a reasonable ask), the clean way on current main is a tiny follow-up: a typed sentinel error (`errUnsupportedIOSVersion`) returned from `StartTunnel`, logged at `Debug` via an `errors.Is` check in `UpdateTunnels` — keeps the backoff working, no fake success. Happy to help wire that up if you want to take it. Really appreciate the contribution. 💙

Issue #318, opened 2023-12-22
ARM64 binary for NPM install to work


serbyxp:
**Is your feature request related to a problem? Please describe.**
Problem installing go-ios from NPM.
I believe there is no binary for the NPM install to complete the postinstall.js script.
It would be nice for home / dev machines running on arm64 (dev / raspberry pi boards) that are running local / containerized node.js to just build without extra steps (installing go etc..)

**Describe the solution you'd like**
NPM install that works on ARM64

**Describe alternatives you've considered**
--install go on arm64 device-- ref. https://go.dev/doc/install
`wget https://go.dev/dl/go1.21.5.linux-arm64.tar.gz`
`rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.5.linux-amd64.tar.gz`
`git clone https://github.com/danielpaulus/go-ios.git,`
`go build` | `go run main.go`
...

**Additional context**
 sudo npm install -g go-ios

npm ERR! code 1

npm ERR! path /usr/lib/node_modules/go-ios

npm ERR! command failed

npm ERR! command sh -c node postinstall.js install

npm ERR! Copying the relevant binary for your platform linux

npm ERR! Error: Command failed: cp ./dist/go-ios-linux-undefined_linux_undefined/ios ./bin/ios

npm ERR! cp: cannot stat './dist/go-ios-linux-undefined_linux_undefined/ios': No such file or directory

npm ERR! 

npm ERR!     at ChildProcess.exithandler (node:child_process:422:12)

npm ERR!     at ChildProcess.emit (node:events:514:28)

npm ERR!     at maybeClose (node:internal/child_process:1105:16)

npm ERR!     at Socket.<anonymous> (node:internal/child_process:457:11)

npm ERR!     at Socket.emit (node:events:514:28)

npm ERR!     at Pipe.<anonymous> (node:net:337:12) {

npm ERR!   code: 1,

npm ERR!   killed: false,

npm ERR!   signal: null,

**npm ERR!   cmd: 'cp ./dist/go-ios-linux-undefined_linux_undefined/ios ./bin/ios'**

npm ERR! }

**npm ERR! Downloaded binary does not contain the binary specified in configuration - ios**

npm ERR! A complete log of this run can be found in: /root/.npm/_logs/2023-12-22T00_28_25_877Z-debug-0.log

muvaf:
@danielpaulus Just opened a PR to fix this issue https://github.com/danielpaulus/go-ios/pull/527

PR #527, opened 2024-12-09
builds: add linux-arm64 binary to the npm package

muvaf:
Mostly replicating what is done for macOS arm64 build; build with GOARCH flag and use `lipo` to build a fat binary. Needs a run in GH actions to confirm it's all working.

Fixes https://github.com/danielpaulus/go-ios/issues/318

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Build
         run: |
-          ((Get-Content -path main.go -Raw) -replace "local-build","${{ steps.create_release.outputs.current_tag }}") | Set-Content -Path main.go  
+          ((Get-Content -path main.go -Raw) -replace "local-build","${{ steps.create_release.outputs.current_tag }}") | Set-Content -Path main.go
           mkdir bin
           go build -ldflags="-s -w" -o bin/ios.exe
           "${{ steps.create_release.outputs.current_tag }}" | Out-File -Encoding utf8NoBOM release_tag -NoNewline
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -143,11 +148,13 @@ jobs:
           mkdir ./npm_publish/dist/go-ios-darwin-amd64_darwin_amd64
           mkdir ./npm_publish/dist/go-ios-darwin-arm64_darwin_arm64
           mkdir ./npm_publish/dist/go-ios-linux-amd64_linux_amd64
+          mkdir ./npm_publish/dist/go-ios-linux-arm64_linux_arm64
           mkdir ./npm_publish/dist/go-ios-windows-amd64_windows_amd64
           cp ./mac-bin/ios ./npm_publish/dist/go-ios-darwin-amd64_darwin_amd64/ios
           cp ./mac-bin/ios ./npm_publish/dist/go-ios-darwin-arm64_darwin_arm64/ios
           cp ./win-bin/ios.exe ./npm_publish/dist/go-ios-windows-amd64_windows_amd64/ios.exe
           cp ./bin/ios ./npm_publish/dist/go-ios-linux-amd64_linux_amd64/ios
```
aluedeke, line 156: can't we use here the platform specific ones: `bin/ios-amd64 bin/ios-arm64`

muvaf: Yes but I didn't want to do a big change. Using the fat binary for both platforms is the original implementation for macOS, so just followed that.

aluedeke (approved):

PR #530, opened 2024-12-09
postinstall: add arm64 as valid arch

muvaf:
https://github.com/danielpaulus/go-ios/pull/529 now publishes binaries for both `linux/amd64` and `linux/arm64` but turns out the postinstallation scripts had two different architecture maps and one was missing `arm64`


aluedeke (approved):

muvaf:
Confirmed on both `linux/amd64` and `linux/arm64`

PR #533, opened 2024-12-13
fix running wda on ios 14..16

chernyka:
As for now, one of the steps to run xcuitest it to upload test configuration to a device. Later a path to uploaded file is just used as `XCTestConfigurationFilePath` env variable for test runner process. We use house_arrest service to upload file to application's container, however when it comes to test runner application built by Xcode it fails with `InstallationLookupFailed` error as runner doesn't have `UIFileSharingEnabled` permission.

I've just tried to run tests using Xcode and looked through env variables in runtime. It appeared that Xcode pass this env indeed but leaves it empty, tested on iOS15.8.3/16.1/17.6.1, Xcode15.3/16.1.

I think we should setup test configuration similar way on iOS17+ or lower, that is not to upload `.xctestconfiguration` as Xcode does not. Unfortunately I don't have devices with earlier OS on board and cannot test it with iOS13 and earlier, so decided not to change current behavior.

### DIFF
--- a/ios/testmanagerd/xcuitestrunner.go
+++ b/ios/testmanagerd/xcuitestrunner.go
@@ -289,35 +289,11 @@ func runXUITestWithBundleIdsXcode15Ctx(
 	}
 	defer conn2.Close()
 
-	installationProxy, err := installationproxy.New(config.Device)
-	if err != nil {
-		return make([]TestSuite, 0), fmt.Errorf("runXUITestWithBundleIdsXcode15Ctx: cannot connect to installation proxy: %w", err)
-	}
-	defer installationProxy.Close()
-	apps, err := installationProxy.BrowseUserApps()
-	if err != nil {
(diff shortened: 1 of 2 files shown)


danielpaulus:
Landed! 🎉 This is re-authored on current main in #790 (merged), with full credit to you via `Co-Authored-By`.

Your fix idea was exactly right — the on-device `.xctestconfiguration` serialization was the breakage, and delivering the config in memory over DTX (the way the iOS 17+ path already works) is the clean fix. I re-implemented it on current main because the branch had drifted (logrus→golog migration + the house_arrest→afc refactor), dropped the unused `setupXcuiTest12`, and kept the change scoped to just the iOS 14–16 path so iOS 17+ stays untouched.

Closing this in favor of #790. Thank you for the fix and for your patience while it sat — really appreciate it. 💙

PR #540, opened 2024-12-26
Set keepaliveidle time more aggressive against connection drops

stoktamisoglu:
Keepaliveidle time should be set more aggresive than the gVisor's 2 hours default idle time. Many NAT and firewalls drop the idle connections more aggressively (30-60 secs apparently).

Keepaliveinterval option doesn't take affect as the keep alive probe doesn't start before the idle time reaches. It is 2 hours by default and before it reaches NAT/Firewall close the connection unlike kernel mode TUN device,

With this change the idle time will be more aggressive and according to several tests this fix connection drops. Long running test more than 24 hours achieved with this fix.

### DIFF
--- a/ios/tunnel/userspace_tunnel.go
+++ b/ios/tunnel/userspace_tunnel.go
@@ -46,9 +46,15 @@ func (iface *UserSpaceTUNInterface) TunnelRWCThroughInterface(localPort uint16, 
 	if err != nil {
 		return fmt.Errorf("TunnelRWCThroughInterface: NewEndpoint failed: %+v", err)
 	}
+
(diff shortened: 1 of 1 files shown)


stoktamisoglu:
Hi, @danielpaulus , @dmissmann, happy new year!, I hope you had a nice festive time. When you have a chance , would you please look at the #540. 

I've tested it in different network environments and it all run  more than day(s) without ant WDA drop. Thanks in advance.

dmissmann (approved):
Don't know too much about the userspace tunnel themselves tbh, but 30s doesn't sound too bad and as you said, it doesn't do anything until it's really idle

Issue #543, opened 2025-01-06
Persistent ios commands like runwda, forward with agent mode enabled terminates the tunnel 


stoktamisoglu:
**Describe the bug**
When you enable experimental agent with ENABLE_GO_IOS_AGENT=yes it starts a kernel tunnel before the command runs for 17+ devices. The tunnel should start and stay after the command though it is a persistent command. 

**To Reproduce**
1. Set environment variable ENABLE_GO_IOS_AGENT=yes
2. Run any command like **ios list** to start an tunnel agent
3. Check if **ios tunnel start** exist in process list after the command ends running
4. After the command finishes run a persistent command like **ios runwda** 
5. After WDA started press Ctrl+C to end the command.
6. **ios tunnel start** process is also terminated

**Expected behavior**
**ios tunnel start** process should stay after the command ends

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows, Linux, Mac
 - Browser N/A
 - Version N/A

**Smartphone (please complete the following information):**
 - Device: Any device with iOS 17+
 - OS: iOS 17+
 - Browser N/A
 - Version N/A

**Additional context**
Tunnel persistency is essential for multi device test automation stability. 
PS: Working on a fix.

danielpaulus:
Resolved on current releases: agent mode now spawns the tunnel agent and blocks on its `/ready` endpoint, which only returns 200 after the TunnelManager's first tunnel update has completed — so runwda/forward and friends no longer race the tunnel and "failed to get tunnel info" is gone from that path. Note the env values changed with #542/#544: use `ENABLE_GO_IOS_AGENT=user` or `ENABLE_GO_IOS_AGENT=kernel` (`yes` is no longer a valid mode). Reopen if you still see the race on the latest release.

_(part of an automated backlog sweep — comment or reopen if this got it wrong)_

PR #544, opened 2025-01-08
Tunnel mode selection feature for agent mode.

stoktamisoglu:
With this change ENABLE_GO_IOS_AGENT can be set to "user" or "kernel" to start selective mode tunnel.

### DIFF
--- a/ios/tunnel/tunnel_api.go
+++ b/ios/tunnel/tunnel_api.go
@@ -54,7 +54,7 @@ func WaitUntilAgentReady() bool {
 	}
 }
 
-func RunAgent(args ...string) error {
+func RunAgent(mode string, args ...string) error {
 	if IsAgentRunning() {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


shamanec (approved):

stoktamisoglu:
Hi @danielpaulus , @dmissmann , would you please look at it at your convivences? I'll submit another PR which is ready for the issue #543 . Thx.

ios/tunnel/tunnel_api.go
```diff
@@ -64,8 +64,18 @@ func RunAgent(args ...string) error {
...
-	cmd := exec.Command(ex, append([]string{"tunnel", "start"}, args...)...)
+	var cmd *exec.Cmd
+
+	if mode == "kernel" {
+		cmd = exec.Command(ex, append([]string{"tunnel", "start"}, args...)...)
+	}
+
+	if mode == "user" {
+		cmd = exec.Command(ex, append([]string{"tunnel", "start", "--userspace"}, args...)...)
+	}
```
dmissmann, line 75: lets verify that we get one of the expected parameters, otherwise it would panic
```suggestion
	var cmd *exec.Cmd
	switch mode {
	case "kernel":
		cmd = exec.Command(ex, append([]string{"tunnel", "start"}, args...)...)
	case "user":
		cmd = exec.Command(ex, append([]string{"tunnel", "start", "--userspace"}, args...)...)
	default:
		return fmt.Errorf("RunAgent: unknown mode: %s. Only 'kernel' and 'user' are supported", mode)
	}
```

dmissmann (changes requested):

ios/tunnel/tunnel_api.go
stoktamisoglu: I've modified with the suggestion. Initial code implemented like the old code for checking only 'yes' value and if only for that value it was calling RunAgent(). Similar, initial code was calling RunAgent() only for 'kernel' and 'user' values. However what @dmissmann suggested is better practice to handle panic independent of the caller. Modified accoringly.

dmissmann (approved):

PR #546, opened 2025-01-13
Reset accessibility settings to default

kvs-coder:
When messing up with gigantic fonts or reduced movement (or other accessibility settings):

Bildschirmfoto 2025-01-09 um 8 18 49 AM (1) (1)

one needs to manually fix it, there is though a method `- (void)deviceResetToDefaultAccessibilitySettings;` which can be called to do the clean-ups


ios/accessibility/accessibility.go
```diff
@@ -14,6 +14,5 @@ func New(device ios.DeviceEntry) (ControlInterface, error) {
 		return ControlInterface{}, err
 	}
 	control := ControlInterface{conn.GlobalChannel()}
-	err = control.init()
```
dmissmann, line 17: does this break the call to `deviceResetToDefaultAccessibilitySettings` if this happens here?

kvs-coder: yes, I noticed that the `init()` panics here all the time https://github.com/danielpaulus/go-ios/pull/546/files#diff-d0706263521565437e277a55d7e4feb44b794334cb4b3bb64de0afbead4a2656R69

kvs-coder: in any case I am introducing breaking changes here

dmissmann: > in any case I am introducing breaking changes here

we could avoid that by adding a `NewWithoutEventChangeListeners` instead and keep the `New` as it is

ios/accessibility/accessibility.go
```diff
@@ -7,6 +7,16 @@ import (
 
 const serviceName string = "com.apple.accessibility.axAuditDaemon.remoteserver"
 
+// NewWithoutEventChangeListeners creates and connects to the given device, a new ControlInterface instance
```
isevendays-saucelabs, line 10: @kvs-coder I would specify why do we need this code, so it is clear when to use this method and why we don't need event change listeners

isevendays-saucelabs: are there any issues if we don't use eventChangeListeners?

kvs-coder: yes, https://github.com/danielpaulus/go-ios/pull/546#discussion_r1914752813 , with the current `init` implementation it should fail, if something goes wrong. For a call `deviceResetToDefaultAccessibilitySettings` we don't need any listener setup, just `fire and forget`

diegoperini (approved):

PR #553, opened 2025-01-24
Refactor how xctestrun files are parsed

si-net:
By introducing new methods and introducing a new metadata object to extract the version, the parsing becomes easier to read and extend in the future.

### DIFF
--- a/ios/testmanagerd/xctestrunnerutils_test.go
+++ b/ios/testmanagerd/xctestrunnerutils_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Helper function to create mock data and parse the .xctestrun file
-func createAndParseXCTestRunFile(t *testing.T) xCTestRunData {
+func createAndParseXCTestRunFile(t *testing.T) schemeData {
 	// Arrange: Create a temporary .xctestrun file with mock data
 	tempFile, err := os.CreateTemp("", "testfile*.xctestrun")
 	assert.NoError(t, err, "Failed to create temp file")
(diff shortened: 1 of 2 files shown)


ios/testmanagerd/xctestrunnerutils_test.go
```diff
@@ -184,11 +184,6 @@ func TestSkipTestIdentifiers(t *testing.T) {
 	}, xcTestRunData.TestConfig.SkipTestIdentifiers, "SkipTestIdentifiers mismatch")
 }
 
-func TestFormatVersion(t *testing.T) {
```
si-net, line 187: There is another test to ensure only version 1 is supported.

ios/testmanagerd/xctestrunutils.go
```diff
@@ -88,57 +81,57 @@ func parseFile(filePath string) (xCTestRunData, error) {
...
-	// Direct decoding of metadata to avoid additional conversion
-	switch v := metadataMap["FormatVersion"].(type) {
-	case int:
-		result.XCTestRunMetadata.FormatVersion = v
-	case uint64:
-		result.XCTestRunMetadata.FormatVersion = int(v)
-	default:
-		return xCTestRunData{}, fmt.Errorf("unexpected FormatVersion type: %T", metadataMap["FormatVersion"])
-	}
+	return xCTestRunData{}, nil
```
dmissmann, line 105: the convention is usually to just return early and align the happy path on the left
```
	if version != 1 {
		return xCTestRunData{}, fmt.Errorf("go-ios currently only supports .xctestrun files in formatVersion 1: "+
			"The formatVersion of your xctestrun file is %d, feel free to open an issue in https://github.com/danielpaulus/go-ios/issues to "+
			"add support", version)
	}

	return parseVersion1(xctestrunFileContent)
```

but since you also plan to add format version 2, you can also do

```
	switch version {
	case 1:
		return parseVersion1(xctestrunFileContent)
	case 2:
		// TODO
		return xCTestRunData{}, fmt.Errorf(".xctestrun format version 2 is not implemented yet")
	default:
		return xCTestRunData{}, fmt.Errorf(".xctestrun format version %d is not supported", version)
	}
```
this also removes the hint to open an issue on github.com, I think saying that's it's not supported is clear enough and isn't too verbose 😉

ios/testmanagerd/xctestrunutils.go
```diff
@@ -25,8 +24,7 @@ import (
 
 // xCTestRunData represents the structure of an .xctestrun file
 type xCTestRunData struct {
-	TestConfig        schemeData        `plist:"-"`
-	XCTestRunMetadata xCTestRunMetadata `plist:"__xctestrun_metadata__"`
+	TestConfig schemeData `plist:"-"`
```
dmissmann, line 27: This is not really used for plist parsing anymore, isn't it? And since `schemeData` is the only field left, can't we just get rid of this struct and use `TestConfig` instead?

dmissmann (approved):

PR #555, opened 2025-01-31
Fsync container

aluedeke:
as i don't know how to add code to someone elses MR here is the cleanup for:

https://github.com/danielpaulus/go-ios/pull/280


danielpaulus (approved):
super nice!

PR #559, opened 2025-02-07
Initial Support for .xctestrun File Format Version 2

bahrimootaz:
**Description :**
Enable the execution of XCTest cases using `.xctestrun` configuration files in **version 2 format**, expanding compatibility and aligning with latest Xcode standards.

**Context :** 
Currently, go-ios supports **.xctestrun** files with **format version 1.** However, Xcode has introduced version 2 of the `.xctestrun` file format, which brings structural changes and enhancements. To maintain compatibility with modern XCTest setups and improve usability for customers, we need to support this newer format.

Reference: new structure of the `.xctestrun` is documented [here](https://keith.github.io/xcode-man-pages/xcodebuild.xctestrun.5.html)

**Testing :**
In this MR we test did test the following scenarios : 
- contains only a single XCTest 🟢 
- contains multiple XCTest targeting the same app 🟢  
- contains multiple XCTest targeting different apps 🟢


ios/testmanagerd/xcuitestrunner.go
```diff
@@ -250,19 +250,29 @@ type TestConfig struct {
...
+			return nil, fmt.Errorf("building test config at index %d: %w", i, err)
+		}
+		xcTestTargets[i] = tc
 	}
 
-	return RunTestWithConfig(ctx, testConfig)
+	var results []TestSuite
+	for _, target := range xcTestTargets {
+		suites, err := RunTestWithConfig(ctx, target)
+		if err != nil {
```
rillgen-saucelabs, line 269: I think this would effectivelly stop test executions if one has an error. I don't know if that's a good idea.

dmissmann: if there are two test targets and the first one returns an error here it won't execute the second one. So it would stop the execution.

I guess we can return both, the partial results and the error

```
var results []TestSuite
var targetErrors []error
for _, target := range xcTestTargets {
	suites, err := RunTestWithConfig(ctx, target)
	if err != nil {
		targetErrors = append(targetErrors, err)
		continue
	}
	...
}
return results, errors.join(targetErrors...)

File: ios/testmanagerd/xctestrunutils.go [hunk 2]
```diff
@@ -25,32 +27,53 @@ import (
 // xCTestRunData represents the structure of an .xctestrun file
 
 // schemeData represents the structure of a scheme-specific test configuration
+type xCTestRunVersion2 struct {
```
si-net, line 30: We should declare this in the method that parses format version 2.

ios/testmanagerd/xctestrunutils.go
```diff
@@ -25,32 +27,53 @@ import (
...
+	IsUITestBundle                  bool
+	CommandLineArguments            []string
+	EnvironmentVariables            map[string]any
+	TestingEnvironmentVariables     map[string]any
+	UITargetAppEnvironmentVariables map[string]any
+	UITargetAppPath                 string
 }
 
-func (data schemeData) buildTestConfig(device ios.DeviceEntry, listener *TestListener) (TestConfig, error) {
+func (data schemeData) buildTestConfig(device ios.DeviceEntry, listener *TestListener, allAps []installationproxy.AppInfo) (TestConfig, error) {
```
si-net, line 52: Note for myself: this smells, there should be a better solution for this.

si-net: In any case, we could skip XCUITests in this MR and implement this as a next step

ios/testmanagerd/xctestrunutils.go
```diff
@@ -116,12 +139,12 @@ func getFormatVersion(xctestrunFileContent []byte) (int, error) {
 	return metadata.Metadata.Version, nil
 }
 
-func parseVersion1(content []byte) (schemeData, error) {
+func parseXCTestRunFileFormatVersion1(content []byte) ([]schemeData, error) {
```
si-net, line 142:
```suggestion
func parseVersion1(xctestrunFile []byte) ([]schemeData, error) {
```

ios/testmanagerd/xctestrunutils.go
```diff
@@ -141,15 +164,46 @@ func parseVersion1(content []byte) (schemeData, error) {
...
+
+	// Check if TestConfigurations is empty
+	if len(testConfigs.TestConfigurations) == 0 {
+		return []schemeData{}, fmt.Errorf("no TestConfigurations found in XCTestRun file(format version: 2); cannot proceed with test parsing")
+	}
+
+	// Check if TestConfigurations contains more than one item
+	if len(testConfigs.TestConfigurations) > 1 {
+		return []schemeData{}, fmt.Errorf("expected exactly one TestConfiguration in XCTestRun file(format version: 2), but found %d", len(testConfigs.TestConfigurations))
+	}
```
si-net, line 194:
```suggestion
	if len(testConfigs.TestConfigurations) != 1 {
		return []schemeData{}, fmt.Errorf("The .xctestrun file you provided contained %d entries in the TestConfiguration list. This list should contain exactly 1 entry. Please revisit your test configuration so that it only contains one entry.", len(testConfigs.TestConfigurations))
	}
```

ios/testmanagerd/xctestrunutils.go
```diff
@@ -25,32 +27,53 @@ import (
...
+type xCTestRunVersion2 struct {
+	ContainerInfo struct {
+		ContainerName string `plist:"ContainerName"`
+	} `plist:"ContainerInfo"`
+	TestConfigurations []struct {
+		TestTargets []schemeData `plist:"TestTargets"`
+	} `plist:"TestConfigurations"`
+}
+
 type schemeData struct {
```
si-net, line 39: This should be called `testTargetConfig` can be done in new MR.

si-net (changes requested):

ios/testmanagerd/xctestrunutils.go
bahrimootaz: True this struct is relevant only in parsing of v2 method. there is no need to make it available in the whole package.

ios/testmanagerd/xctestrunutils.go
bahrimootaz: i think it makes no difference, needed to fully support v2, running xcuitests is just as important as xctests, also for xcuitests we don't always require a bundleId since in some cases the ui test can trigger system apps (in this case the `UITargetAppPath` will be empty). that point was considered while working on this and it's error safe code, if we face any issue getting the the BundleID we let the execution proceed and return a partial the errors combined with the results. but most probably we shouldn't get errors.

ios/testmanagerd/xcuitestrunner.go
bahrimootaz: @dmissmann nice idea, we implemented this

ios/testmanagerd/xctestrunutils_test.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	mockListener := &TestListener{}
+
+	// Act: Convert XCTestRunData to TestConfig
+	testConfig, err := xcUITestTarget.buildTestConfig(mockDevice, mockListener)
+
+	// Assert: Validate the returned TestConfig
+	assert.NoError(t, err, "Error converting to TestConfig")
+
+	return testConfig, mockDevice, mockListener
```
dmissmann, line 329: `mockDevice` and `mockListener` are always ignored, we can remove them

ios/testmanagerd/xctestrunutils_test.go
```diff
@@ -0,0 +1,363 @@
+package testmanagerd
```
dmissmann, line 1: The number of tests can be reduces significantly. There are nine tests using `parseXCTestRunFileVersion1`, and that can simply be one. Initialize the `schemeData` struct to what you expect and just assert against that. The test output should show you the differences in a easy to read format which also makes the assertion messages like `"TestHostBundleIdentifier mismatch"` obsolete

dmissmann (changes requested):

si-net:
We discovered another issue with how we accumulate the test suites, we will have to revisit this today.

ios/testmanagerd/xctestrunutils_test.go
bahrimootaz: done

ios/testmanagerd/xctestrunutils_test.go
bahrimootaz: we did restructured the testing file, can you please check again.

dmissmann (approved):

PR #560, opened 2025-02-09
Persistent go-ios agent after the initial go-ios terminates.

stoktamisoglu:
When go-ios is started in agent mode, the agent terminates when the parent go-ios process terminates. If you run go-ios in multiple terminal for multiple device test automation the this can disrupt other terminal sessions which may be using the first started tunnel.

With this  PR we use SysProcAttr for *nix and Window platforms.

For *nix platforms setting Setsid: true starts agent in an independent new *nix session. And for windows as this setting is not available by OS nature we can use CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP to start the process in a new process group.

They are tested on *nix and Windows and works well.


stoktamisoglu:
Hi @dmissmann and @danielpaulus , would you review the #560 at your conveniences?

dmissmann (approved):
That makes sense, but do I get it right that now the only way to kill the agent in the background is with a `kill ...` command?

stoktamisoglu:
> That makes sense, but do I get it right that now the only way to kill the agent in the background is with a `kill ...` command?

Thanks for the review @dmissmann.

Formal and graceful way to terminate the tunnel is still 
`go-ios tunnel stopagent`

Of course, kill and and the signals are still alternative to stop the tunnel.

PR #563, opened 2025-02-11
Enable Full Support for Multiple XCTest/XCUITest Cases configured in xctestrun File

bahrimootaz:
**Description**
This PR extends the support for xctestrun files with format version 2, building on the initial implementation merged in [#559](https://github.com/danielpaulus/go-ios/pull/559).

With this update, we enable full support for XCUITests by utilizing the parsed test configurations from an xctestrun file. 

**Testing :**
The following scenarios were tested to validate the changes:

✅ Single XCTest case
✅ Single XCUITest case
✅ Multiple XCTest cases targeting the same app
✅ Multiple XCTest cases targeting different apps
✅ Multiple XCUITest cases targeting the expected app
✅ Mixed XCUITest and XCTest cases targeting the expected app
✅ Mixed XCUITest and XCTest cases targeting different apps

These enhancements ensure that `go-ios` can reliably handle diverse test configurations parsed from an `xctestrun` file in both format versions.

### DIFF
--- a/ios/testmanagerd/xctestrunutils.go
+++ b/ios/testmanagerd/xctestrunutils.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"github.com/danielpaulus/go-ios/ios"
+	"github.com/danielpaulus/go-ios/ios/installationproxy"
 	"howett.net/plist"
 	"io"
 	"maps"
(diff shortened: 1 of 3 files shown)


ios/testmanagerd/xctestrunutils.go
```diff
@@ -21,32 +22,43 @@ import (
...
+	IsUITestBundle                  bool
+	CommandLineArguments            []string
+	EnvironmentVariables            map[string]any
+	TestingEnvironmentVariables     map[string]any
+	UITargetAppEnvironmentVariables map[string]any
+	UITargetAppPath                 string
 }
 
-func (data schemeData) buildTestConfig(device ios.DeviceEntry, listener *TestListener) (TestConfig, error) {
+func (data schemeData) buildTestConfig(device ios.DeviceEntry, listener *TestListener, allAps []installationproxy.AppInfo) (TestConfig, error) {
```
dmissmann, line 37: lets name it `allApps` or `installedApps`

ios/testmanagerd/xctestrunutils.go
```diff
@@ -21,32 +22,43 @@ import (
...
 
 	if data.IsUITestBundle {
 		maps.Copy(testEnv, data.EnvironmentVariables)
 		maps.Copy(testEnv, data.TestingEnvironmentVariables)
+		maps.Copy(testEnv, data.UITargetAppEnvironmentVariables)
+		// Only call getBundleID if :
+		// - allAps is provided
+		// - UITargetAppPath is populated since it can be empty for UI tests in some edge cases
+		if len(data.UITargetAppPath) > 0 && allAps != nil {
+			bundleId = *getBundleID(allAps, data.UITargetAppPath)
```
dmissmann, line 52: lets not return a pointer from `getBundleID` but `""` instead of `nil`. If we get `nil` here this panics, and if we get `""` we get the same behaviour as when we never end up at this statement

dmissmann (changes requested):

ios/testmanagerd/xctestrunutils.go
bahrimootaz: done

ios/testmanagerd/xctestrunutils.go
bahrimootaz: We changed `getBundleId` to return an empty string instead of a pointer on a string.

ios/testmanagerd/xcuitestrunner.go
```diff
@@ -256,14 +256,12 @@ func StartXCTestWithConfig(ctx context.Context, xctestrunFilePath string, device
 			fmt.Errorf("error parsing xctestrun file: %w", err),
 		}
 	}
+	svc, _ := installationproxy.New(device)
+	installedApps, _ := svc.BrowseUserApps()
```
dmissmann, line 260: are the errors ignored intentionally here?
Also, `svc` needs to be closed after we're done

dmissmann (approved):

dmissmann (approved):

PR #577, opened 2025-03-18
fix: recognizing "export USBMUXD_SOCKET_ADDRESS=UNIX:///tmp/usbmuxd" as tcp connections.

Ylarod:
Enhance GetUsbmuxdSocket to support various socket address formats, fix recognizing "export USBMUXD_SOCKET_ADDRESS=UNIX:///tmp/usbmuxd" as tcp connections.


ios/usbmuxconnection.go
```diff
@@ -32,11 +32,19 @@ func ToUnixSocketPath(socketAddress string) string {
...
+		// if it has a scheme, just use it
+		if strings.Contains(socket_override, "://") {
+			return socket_override
+		}
+		// if it is a file, use it as unix socket
+		if _, err := os.Stat(socket_override); err == nil {
 			return "unix://" + socket_override
 		}
+		// if it is a tcp address, use it
+		if strings.Contains(socket_override, ":") && strings.Contains(socket_override, ".") {
```
aluedeke: how about `hostname:port` this?

Can't we fix this and maintain backward compatibility?

Check if there is a scheme on the string if so return it. If not fallback to the old logic.

danielpaulus: Good call — addressed in a maintainer commit (94daa3f). Reworked it to keep backward compatibility instead of panicking:

- If the value has a scheme (\`://\`), use it as-is, but lower-case the scheme so \`net.Dial\` accepts it (it rejects \`UNIX\`) — that's what makes \`USBMUXD_SOCKET_ADDRESS=UNIX:///tmp/usbmuxd\` work end to end.
- Otherwise fall back to the original logic: a \`:\` means tcp \`host:port\` (so \`hostname:port\` like \`localhost:27015\` still works), anything else is a unix socket path.
- Dropped the \`os.Stat\` probe and the \`panic\` on unrecognized input.

Added a table-driven test (\`TestGetUsbmuxdSocket\`) covering scheme/uppercase-scheme/hostname:port/ip:port/bare-path.

danielpaulus (approved):

PR #600, opened 2025-05-31
Fix nil pointer dereference in testlistener.go

iSevenDays:
- Handle cases where testCaseDidStartForClass is called before testSuiteDidStart
- Create new suite automatically if none exists
- Add unit tests to verify fix


iSevenDays:
I've created a simple test case that demonstrates the fix for.
To run the test:
`go run fixed_implementation.go`

The output should confirm that a new test suite is created and the test case is added properly.

```
package main

import (
	"fmt"
	"os"
	"time"
)

type TestCase struct {
	ClassName  string
	MethodName string
}

type TestSuite struct {
	Name      string
	StartDate time.Time
	TestCases []TestCase
}

type TestListener struct {
	TestSuites       []TestSuite
	runningTestSuite *TestSuite
}

func (t *TestListener) findTestSuite(className string) *TestSuite {
	fmt.Printf("Looking for suite with name: %s\n", className)
	if t.runningTestSuite != nil && t.runningTestSuite.Name == className {
		fmt.Println("Found running test suite")
		return t.runningTestSuite
	}
	for _, suite := range t.TestSuites {
		if suite.Name == className {
			fmt.Printf("Found existing test suite: %s\n", suite.Name)
			return &suite
		}
	}
	fmt.Println("No test suite found")
	return nil
}

func (t *TestListener) testCaseDidStartForClass(testClass string, testMethod string) {
	// Find the existing test suite or create a new one if not found
	ts := t.findTestSuite(testClass)
	if ts == nil {
		if t.runningTestSuite != nil {
			ts = t.runningTestSuite
		} else {
			d := time.Now()
			newSuite := TestSuite{
				Name:      testClass,
				StartDate: d,
				TestCases: []TestCase{},
			}
			t.TestSuites = append(t.TestSuites, newSuite)
			ts = &t.TestSuites[len(t.TestSuites)-1]
		}
	}

	// Add the test case to the suite
	fmt.Printf("Adding test case (%s) to suite (%s)\n", testMethod, ts.Name)
	ts.TestCases = append(ts.TestCases, TestCase{
		ClassName:  testClass,
		MethodName: testMethod,
	})
	fmt.Printf("Test cases in suite %s: %d\n", ts.Name, len(ts.TestCases))
}

func main() {
	testListener := &TestListener{}

	fmt.Println("Starting test...")
	testListener.testCaseDidStartForClass("MyTestClass", "testMethod")

	fmt.Printf("Number of test suites: %d\n", len(testListener.TestSuites))
	if len(testListener.TestSuites) > 0 {
		for i, suite := range testListener.TestSuites {
			fmt.Printf("Suite #%d: Name=%s, Number of test cases=%d\n",
				i+1, suite.Name, len(suite.TestCases))
		}
	}

	if len(testListener.TestSuites) == 1 && len(testListener.TestSuites[0].TestCases) == 1 {
		fmt.Println("Test passed: Test case was added to a new test suite")
	} else {
		fmt.Println("Test failed: Could not add test case to a new test suite")
		os.Exit(1)
	}
}
```

ios/testmanagerd/testlistener.go
```diff
@@ -164,7 +164,31 @@ func (t *TestListener) testSuiteDidStart(suiteName string, date string) {
 }
 
 func (t *TestListener) testCaseDidStartForClass(testClass string, testMethod string) {
+	// Find the existing test suite or create a new one if not found
 	ts := t.findTestSuite(testClass)
+	if ts == nil {
+		// If no test suite is found and we're not in a running test suite,
+		// we should use the runningTestSuite instead of creating a new one.
+		// This handles cases where testCaseDidStartForClass is called before
+		// testSuiteDidStart, which would otherwise result in a nil pointer dereference.
```
dmissmann, line 173: > This handles cases where testCaseDidStartForClass is called before testSuiteDidStart, which would otherwise result in a nil pointer dereference.

Are you sure that's happening and the call for testSuiteDidStart` after `testCaseDidStartForClass` is not for a different test suite? It sounds very strange

dmissmann: I meant reproducing on an actual device with a real test execution

PR #601, opened 2025-06-02
support more then 34 devices

codeskyblue:


ios/listdevices.go
```diff
@@ -102,9 +103,47 @@ func (muxConn *UsbMuxConnection) ListDevices() (DeviceList, error) {
 	if err != nil {
 		return DeviceList{}, fmt.Errorf("Failed getting devicelist: %v", err)
 	}
+	muxResponse := MuxResponsefromBytes(response.Payload)
+	if muxResponse.Number == 4294967295 {
```
gmegidish, line 107: it looks like a magic number when put into decimal. maybe use 0xffffffff or a constant?
also, why are we getting this number instead of a device list? can you elaborate please?

ios/listdevices.go
```diff
@@ -102,9 +103,47 @@ func (muxConn *UsbMuxConnection) ListDevices() (DeviceList, error) {
...
+	}()
+	var deviceList = DeviceList{
+		DeviceList: make([]DeviceEntry, 0, 40),
+	}
+	for {
+		select {
+		case entry := <-deviceEntryChan:
+			deviceList.DeviceList = append(deviceList.DeviceList, entry)
+		case <-time.After(100 * time.Millisecond):
+			return deviceList, nil
```
gmegidish, line 142: is 100ms the only way to know we have reached the last device? such absolute timing can be random when dealing with services. what do you think?

ios/listdevices.go
```diff
@@ -102,9 +103,47 @@ func (muxConn *UsbMuxConnection) ListDevices() (DeviceList, error) {
...
+			if err != nil {
+				break
+			}
+			if msg.DeviceAttached() {
+				deviceEntryChan <- msg.DeviceEntry()
+			}
+		}
+	}()
+	var deviceList = DeviceList{
+		DeviceList: make([]DeviceEntry, 0, 40),
```
gmegidish, line 135: why was 34 a limit? and why put 40 as the new limit?

main.go
```diff
@@ -297,6 +297,9 @@ The commands work as following:
 	log.Debug(arguments)
 
 	skipAgent, _ := os.LookupEnv("ENABLE_GO_IOS_AGENT")
+	if skipAgent == "yes" {
```
gmegidish, line 300: how is this change related to the PR?

gmegidish (changes requested):

ios/listdevices.go
codeskyblue: This number should be -1, Bu the Number Type is uint32. So it is 0xffffffff;

ios/listdevices.go
codeskyblue: em.. do you have some better way?

main.go
codeskyblue: ok, this will be removed.

ios/listdevices.go
codeskyblue: In most conditions, phone number is bellow 40 (from my experience).

PR #612, opened 2025-07-24
Fix fsync with --app=bundleId

### LINKED ISSUE #593: "ios fsync --udid=<udid> --app=<bundleID> tree --path=/Documents" throw error "fsync: connect afc service failed"
When run "ios fsync --udid=<udid> --app=<bundleID> tree --path=/Documents", it print an error like
{"err":"InstallationLookupFailed","level":"fatal","msg":"fsync: connect afc service failed","time":"2025-05-07T09:30:58Z"}

For the same device, when run "tidevice -u <udid> fsync -B bundleID ls /Documents", it will print right result.

I have tested two devices:
1. iphone 11, ios 17.6.1
2. iphone 6s plus, ios 15.7.5

zemanj:
Fixes #593.

According to [libimobiledevice issue 193](https://github.com/libimobiledevice/libimobiledevice/issues/193) from 2015 (in particular, [this comment](https://github.com/libimobiledevice/libimobiledevice/issues/193#issuecomment-106424991)), the service `com.apple.mobile.house_arrest` doesn't support the `VendContainer` command anymore (unless the app was installed via XCode).
However, for apps that have [`UIFileSharingEnabled`](https://developer.apple.com/documentation/bundleresources/information-property-list/uifilesharingenabled) set to `true` in their [information property list](https://developer.apple.com/documentation/bundleresources/information-property-list), the `VendDocuments` command can be used instead to access the app's `/Documents` directory.

This PR replaces `VendContainer` by `VendDocuments` to make this work.

Note that when using the go-iOS `fsync` command with the option `--app=BUNDLE_ID` set, the `PATH` passed with the option `--path=PATH` has to start with `/Documents` (in particular, `/` won't work).
For example, to use the `fsync tree` command for a specific app, one has to call `ios fsync --app=BUNDLE_ID tree --path=/Documents`.

### DIFF
--- a/ios/afc/fsync.go
+++ b/ios/afc/fsync.go
@@ -50,24 +50,24 @@ func New(device ios.DeviceEntry) (*Connection, error) {
 	return &Connection{deviceConn: deviceConn}, nil
 }
 
-func NewContainer(device ios.DeviceEntry, bundleID string) (*Connection, error) {
+func NewVendDocuments(device ios.DeviceEntry, bundleID string) (*Connection, error) {
 	deviceConn, err := ios.ConnectToService(device, "com.apple.mobile.house_arrest")
 	if err != nil {
 		return nil, err
 	}
-	err = VendContainer(deviceConn, bundleID)
+	err = VendDocuments(deviceConn, bundleID)
 	if err != nil {
 		return nil, err
 	}
 	return &Connection{deviceConn: deviceConn}, nil
 }
 
-func VendContainer(deviceConn ios.DeviceConnectionInterface, bundleID string) error {
+func VendDocuments(deviceConn ios.DeviceConnectionInterface, bundleID string) error {
 	plistCodec := ios.NewPlistCodec()
-	vendContainer := map[string]interface{}{"Command": "VendContainer", "Identifier": bundleID}
-	msg, err := plistCodec.Encode(vendContainer)
+	vendDocuments := map[string]interface{}{"Command": "VendDocuments", "Identifier": bundleID}
+	msg, err := plistCodec.Encode(vendDocuments)
 	if err != nil {
-		return fmt.Errorf("VendContainer Encoding cannot fail unless the encoder is broken: %v", err)
+		return fmt.Errorf("VendDocuments encoding cannot fail unless the encoder is broken: %v", err)
 	}
 	err = deviceConn.Send(msg)
 	if err != nil {
@@ -81,8 +81,8 @@ func VendContainer(deviceConn ios.DeviceConnectionInterface, bundleID string) er
 	return checkResponse(response)
 }
 
-func checkResponse(vendContainerResponseBytes []byte) error {
-	response, err := plistFromBytes(vendContainerResponseBytes)
+func checkResponse(vendDocumentsResponseBytes []byte) error {
+	response, err := plistFromBytes(vendDocumentsResponseBytes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


danielpaulus:
Thanks for digging into this. I reviewed the current diff and this is not just a rename: the behavior changes the house_arrest request from `VendContainer` to `VendDocuments`. That may be the right mode for some workflows, but it also changes the filesystem root that iOS vends for `fsync --app`.\n\nI do not think we should merge this as-is before we have real-device/e2e coverage for the affected command. The safer direction is probably:\n\n- keep the existing `--app` behavior tied to `VendContainer`, unless proven wrong by device tests;\n- add explicit support for `VendDocuments` as a separate CLI mode/flag if that is what users need;\n- add e2e coverage proving which paths are visible for both modes.\n\nLeaving this note here so we do not lose the finding when we return to the PR.

danielpaulus:
Maintainer review notes — thanks for this, @zemanj, and for the precise libimobiledevice writeup. The diagnosis is correct: `house_arrest` no longer supports `VendContainer` for apps that weren't installed via Xcode, and `VendDocuments` is the right command for normal apps that set `UIFileSharingEnabled` (#593). We want this fixed. Flagging two things before it can land, so it doesn't get merged as-is and cause a regression:

**1. The PR has gone stale (`CONFLICTING`).** Since it was opened, the afc package was refactored. The functions this PR edits — `NewContainer` / `VendContainer` in `ios/afc/fsync.go` — no longer exist on `main`. The vend logic now lives in `ios/house_arrest/house_arrest.go` as a private `vendContainer`, and `house_arrest.New(...)` returns an `*afc.Client`. So the change needs to be re-pointed at `house_arrest.go`.

**2. ⚠️ A blanket swap to `VendDocuments` would break XCUITest running.** `house_arrest.New` is shared by two callers:
- `cmd_device_files.go` — the user-facing `ios file` / `ios fsync --app=<bundleID>` command (what this PR is about), and
- `ios/testmanagerd/xcuitestrunner.go:555` — **the XCUITest runner** (how go-ios launches WebDriverAgent / DeviceKit / any `runtest`).

The runner writes its test config to `tmp/<uuid>.xctestconfiguration` (`createTestConfigOnDevice`) — i.e. the app-container root, **not** `/Documents`. `VendDocuments` only vends `/Documents`, so switching the shared `New` to `VendDocuments` would make the runner's `Open("tmp/...")` fail and break WDA/DeviceKit/`runtest` entirely. (The runner apps are Xcode/dev-installed, so `VendContainer` keeps working for them — they must stay on it.)

**Recommended approach (surgical, no regression):**
- Keep `house_arrest.New` on `VendContainer` (used by testmanagerd — must not change).
- Add a `house_arrest.NewDocuments(device, bundleID)` that vends via `VendDocuments` (factor the shared `vend`/`checkResponse`/`plistFromBytes` to take the command string).
- In `runFsyncCommand` (and the `ios file` `--app` path), **try `house_arrest.New` first, and on failure fall back to `house_arrest.NewDocuments`** — dev apps keep full-container access, normal `UIFileSharingEnabled` apps get `/Documents`. After a failed `VendContainer` the service closes the connection, so the fallback opens a fresh `ConnectToService`.
- Note in `--app` help text that for normal apps paths are rooted at `/Documents` (e.g. `ios fsync --app=BUNDLE_ID tree --path=/Documents`), matching what you documented here.

Your core insight (`VendDocuments` for file-sharing apps) is exactly right and will be preserved with credit when this is ported onto current `main`. Parking it here for now rather than merging a stale/regressing diff. 🙏

danielpaulus:
Thank you so much for this, @zemanj — and sorry it sat so long. 🙏 You spotted a real problem: `VendContainer` is refused for apps that only expose their `Documents/` via `UIFileSharingEnabled`, so `ios fsync --app=<id>` is dead-on-arrival for those. That's a genuine bug and your instinct to reach for `VendDocuments` was right.

I'm closing this one only because the shape and the base moved out from under it:
- Swapping the **default** wholesale would regress everyone who relies on full-container access (`Library/`, `Caches`, `tmp/`), and — critically — it would break `ios runtest`, which writes its `.xctestconfiguration` into the container's `tmp/` via house_arrest (unreachable when the AFC root is `Documents/`).
- The reference tools agree on offering **both**, container-default: libimobiledevice `afcclient` has `--container`/`--documents`, and pymobiledevice3 defaults `documents_only=False`.
- The afc package was also refactored since (the exported `VendContainer`/`NewContainer` no longer exist; the vend logic now lives in `ios/house_arrest/house_arrest.go`), so the diff no longer applies.

The ideal, and I'd genuinely love a PR for it: keep `VendContainer` as the default and add an opt-in `--documents` flag on `fsync`/`file` (threaded through `house_arrest.New`), plus an e2e test against an app with `UIFileSharingEnabled`. That captures exactly the case you hit without taking anything away. Thanks again for surfacing this — it's a real gap and the fix is wanted. 💙

PR #627, opened 2025-10-02
Refactor AFC

dmissmann:
The current implementation has **a lot** of repeated code. Every method that communicates with the afc service on the device has it's own logic for building the package and maintaining the package identifiers.

This is now done in a single place that abstracts away of this logic that was repeated over and over before.

Also, file I/O works now similar to "regular" filesystems. When we open a file using the afc service, we get back a `io.ReadWriteCloser`, which allows us to use functions like `io.Copy` on those files.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,36 @@
+#!/bin/sh
+# Pre-commit hook for go-ios
+# Runs gofmt and golangci-lint before allowing commits
+
+set -e
+
+echo "Running pre-commit checks..."
+
+# Check if golangci-lint is installed
+if ! command -v golangci-lint &> /dev/null; then
+    echo "Error: golangci-lint is not installed."
+    echo "Install it with: brew install golangci-lint"
+    echo "Or see: https://golangci-lint.run/docs/welcome/install/"
+    exit 1
+fi
+
+# Check formatting
+echo "Checking code formatting..."
+UNFORMATTED=$(gofmt -l .)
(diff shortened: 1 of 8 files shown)


dmissmann:
> do I understand correctly that fsync --pull and --push only operate on single files now? maybe in main.go:1136 we could use smth like
> 
> ```
>     _, err = io.Copy(local, remote)
>     if dp == "" {
>         dp = "."
> }
> err = afcService.Pull(sp, dp)
> ```
> 
> instead of
> 
> ```
> local, err := os.Create(path.Join(dp, filepath.Base(sp)))
> exitIfError("failed to open local file", err)
> defer local.Close()
> remote, err := afcService.Open(sp, afc.READ_ONLY)
> ```
> 
> I think the pulling/pushing of directories may fail with AFC

I reverted the changes in `main.go` now and it works as before. And that was pulling works with directories as well, but pushing was always file-only

ios/afc/errors.go
```diff
@@ -0,0 +1,118 @@
...
+	case errMuxError:
+		return errors.New("MuxError")
+	case errNoMemory:
+		return errors.New("NoMemory")
+	case errNotEnoughData:
+		return errors.New("NotEnoughData")
+	case errDirNotEmpty:
+		return errors.New("DirNotEmpty")
+	default:
+		return nil
```
iSevenDays: @dmissmann maybe we should return an error here?
```
return fmt.Errorf("unknown AFC error code: %d", errorCode)
```

PR #629, opened 2025-10-07
Support new audit api for ios18

sakhisheikh:
**Problem:** 
`init` fails for AX service because deviceAllAuditCaseIDs returns `nil` as response for AuditCaseIDs.

**Solution:**

iOS `api >= 15`  has new method to fetch Audit IDs which is `deviceAllSupportedAuditTypes`. We are implementing this now based on api version.

Note: we need to silently fail `deviceHumanReadableDescriptionForAuditCaseID`  otherwise it blocks enabling AX.

### DIFF
--- a/ios/accessibility/accessibility_control.go
+++ b/ios/accessibility/accessibility_control.go
@@ -64,7 +64,7 @@ func (a ControlInterface) init() error {
 	}
 	log.Info("Api version:", apiVersion)
 
-	auditCaseIds, err := a.deviceAllAuditCaseIDs()
+	auditCaseIds, err := a.deviceAllAuditCaseIDs(apiVersion)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


ios/accessibility/accessibility_control.go
```diff
@@ -299,7 +306,14 @@ func (a ControlInterface) deviceHumanReadableDescriptionForAuditCaseID(auditCase
 	if err != nil {
 		return "", err
 	}
-	return response.Payload[0].(string), nil
+	if response.Payload == nil {
+		return "", fmt.Errorf("no payload in response")
+	}
```
dmissmann, line 311: we also need at least one element in the slice

```suggestion
	if len(response.Payload) == 0 {
		return "", fmt.Errorf("no payload in response")
	}
```

ios/accessibility/accessibility_control.go
```diff
@@ -299,7 +306,14 @@ func (a ControlInterface) deviceHumanReadableDescriptionForAuditCaseID(auditCase
...
+	str, ok := response.Payload[0].(string)
+	if !ok {
+		return "", fmt.Errorf("payload[0] is not a string")
```
dmissmann, line 314:
```suggestion
		return "", fmt.Errorf("unexpected payload type: %T", payload[0])
```

PR #637, opened 2025-10-22
Implement reset accessibility endpoint integration

iSevenDays:
This PR implement a new endpoint to reset iOS accessibility settings.

The changes would allow us to call this new endpoint `resetaccessibility` to reset iOS accessibility settings like font size etc.


danielpaulus (approved):
Good stuff!

PR #639, opened 2025-10-24
Suppress superflous errors generated closing a forwarding connection

briankrznarich:
Just noticed that I've had this change locally on disk for a year.  If you look at the logic for `ConnListener.Close()` , it calls `cl.listener.Close()`, which in turn causes `listener.Accept()` calls in related goroutines to return with errors. These errors are logged to stderr before the routines end.

These errors should not be emitted when the connection is manually/cleanly closed. The existing call to `close(cl.quit)` has no impact on preventing these errors.

Especially when using `go-ios` as a library where connections are dynamically opened and closed, these errors are quite undesirable.

### DIFF
--- a/ios/forward/forward.go
+++ b/ios/forward/forward.go
@@ -69,6 +69,13 @@ func connectionAccept(cl *ConnListener, deviceID int, phonePort uint16) {
 		default:
 			clientConn, err := cl.listener.Accept()
 			if err != nil {
+				// Close() shuts the listener down, which unblocks this Accept()
+				// with net.ErrClosed. That's a clean teardown, not a failure, so
+				// don't log it as an error — just exit the accept loop.
+				if errors.Is(err, net.ErrClosed) {
+					golog.Info("closed listener successfully", "module", logModule, "deviceID", deviceID, "phonePort", phonePort)
+					return
+				}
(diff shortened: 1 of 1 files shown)


danielpaulus:
Huge thanks for this, @briankrznarich — and sorry it sat for a year. 🙏 You were dead right: `Close()` tears down the listener, which unblocks `Accept()` and that clean shutdown was being logged as a scary `error accepting new connection` on every single close. Exactly the kind of noise that's miserable when you're embedding go-ios as a library and cycling forwards.

Since you opened this, the package migrated from logrus to our internal `golog` seam, and the companion half of your fix — the `forward: close clientConn ...` lines in `StartNewProxyConnection` — landed separately in #754. So I took you up on "allow edits from maintainers" and rebased your change onto current `main`, keeping just the `connectionAccept` part that #754 didn't cover. I also leaned on the fact that a cleanly-closed listener returns `net.ErrClosed`, so we can detect the clean-shutdown case without adding a `closedCleanly` bool or threading the listener through `StartNewProxyConnection` — same idiom #754 used.

Your authorship is preserved on the commit. gofmt/vet/build all clean. Merging once CI's green. Thanks again for the patience and the sharp diagnosis! 🎉

PR #643, opened 2025-11-12
Accessibility Inspector: return spoken description for focused element

sakhisheikh:
After navigating to a element we would like to return the spoken description of the focused element. That allows a client to consume the string and could potentially feed it to an audio engine.

Created some helper method to avoid repetition.

### DIFF
--- a/ios/accessibility/accessibility_control.go
+++ b/ios/accessibility/accessibility_control.go
@@ -217,8 +217,28 @@ func (a ControlInterface) extractSpokenDescription(innerValue map[string]interfa
 	return ""
 }
 
-// extractPlatformElementBytes extracts the platform element bytes from innerValue.
-// Extraction path: ElementValue_v1 -> Value -> Value -> PlatformElementValue_v1 -> Value ([]byte)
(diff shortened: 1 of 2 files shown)


ios/accessibility/accessibility_control.go
```diff
@@ -185,47 +186,66 @@ func (a ControlInterface) Move(ctx context.Context, direction MoveDirection) (AX
...
+		return desc
 	}
 
-	valMap, ok := axElement["Value"].(map[string]interface{})
-	if !ok {
-		return AXElementData{}, fmt.Errorf("AX element inner \"Value\" is not a map, got %T", axElement["Value"])
+	return ""
+}
+
+// extractPlatformElementBytes extracts the platform element bytes from innerValue.
```
buxomant: Minor quip, but this line doesn't tell us anything that we wouldn't glean from the method name and signature. Maybe if it explained what "platform element" or "inner value" mean (conceptually), otherwise it can probably be removed.

The "extraction path" comment below is useful though.

PR #645, opened 2025-11-21
Support GO_IOS_UDID environment variable

aluedeke:
## Summary

Add support for reading device UDID from the `GO_IOS_UDID` environment variable when the `--udid` flag is not specified.

## Changes

- Added fallback to `GO_IOS_UDID` environment variable in main.go
- The `--udid` flag takes precedence if both are specified

## Usage

Users can now set the device UDID via environment variable:

```bash
export GO_IOS_UDID=00008030-000E04D62E8B802E
ios file ls --app=com.example.app
```

Or continue using the flag as before:

```bash
ios --udid=00008030-000E04D62E8B802E file ls --app=com.example.app
```

## Test plan

- Build succeeds
- Command works with `--udid` flag (existing behavior)
- Command works with `GO_IOS_UDID` env var (new behavior)
- Flag takes precedence over env var when both are set

### DIFF
--- a/main.go
+++ b/main.go
@@ -149,7 +149,7 @@ Options:
   --nojson                  Disable JSON output
   --pretty                  Pretty-print JSON command output
   -h --help                 Show this screen.
-  --udid=<udid>             UDID of the device.
+  --udid=<udid>             UDID of the device. Can also be set via GO_IOS_UDID environment variable.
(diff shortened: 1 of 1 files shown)


gmegidish:
@aluedeke should you add this to README.md as well?

PR #654, opened 2025-12-03
Fix crashreport pattern matching and cwd handling

iSevenDays:
skip directories when walking crash reports and match patterns on the basename e.g. wildcards like */*.ips will catch nested files, use PullSingleFile to avoid opening directories as files


dmissmann (commented):
I'll get that we shouldn't try to open directories, but why stop traversing the completely? There are crash reports inside those subdirectories

PR #664, opened 2025-12-19
Bugfix: use service version instead of OS version for Audit API

sakhisheikh:
**Problem:**

Enable AX inspector breaks for iOS 14 or below because during inspector's initialisation we grab audit case id. We use the newer version api for iOS 14 which breaks enabling Accessibility Inspector.

**Solution:** 
The Accessibility Audit API changed after iOS 15. The AX inspector provides `deviceAPIVersion` which uses a different version numbering scheme than the OS version API version 21 corresponds to iOS 15 and so on.


iSevenDays (approved):
LGTM

PR #682, opened 2026-02-06
fix: prevent socket leak in TunnelManager

vbragaru:
## Problem                                                                                                                                      
                                                                                                                                                  
  The `TunnelManager.UpdateTunnels()` function was creating socket connections to `/var/run/usbmuxd` for every device on every call, including    
  network-connected devices that can never establish tunnels. This caused thousands of orphaned Unix sockets to accumulate over time.             
                                                                                                                                                  
  **Root cause:**                                                                                                                                 
  1. `ListDevices()` returns all devices including 50+ network-connected devices                                                                  
  2. For each device not in `m.tunnels`, it calls `startTunnel()` which creates socket connections via `ios.GetProductVersion()`                  
  3. Network devices always fail to connect, but are **never tracked as failed**                                                                  
  4. Next `UpdateTunnels()` call retries the same failing devices                                                                                 
  5. With default 1-second interval × 50+ network devices = hundreds of leaked sockets per hour                                                   
                                                                                                                                                  
  **Evidence:** After running overnight, `lsof` showed 3000+ orphaned Unix sockets:                                                               
  Busymate 94207    v    5u     unix 0x5bcac11ae229dddf   ->(none)                                                                                
  Busymate 94207    v    6u     unix 0x1c4c7d88a25e315b   ->(none)                                                                                
  ... (3000+ more)                                                                                                                                
                                                                                                                                                  
  ## Solution                                                                                                                                     
                                                                                                                                                  
  1. **Skip network-connected devices** - They cannot establish tunnels anyway (`ConnectionType == "Network"`)                                    
                                                                                                                                                  
  2. **Track failed devices with exponential backoff** - Instead of retrying every cycle:                                                         
     - 30s after 1st failure                                                                                                                      
     - 60s after 2nd failure                                                                                                                      
     - 120s after 3rd failure                                                                                                                     
     - Max 5 minutes between retries                                                                                                              
                                                                                                                                                  
  3. **Clean up on disconnect** - Remove device from failed list when it disconnects, allowing fresh retry on reconnect                           
                                                                                                                                                  
  4. **Clean up on success** - Remove device from failed list when tunnel succeeds                                                                
                                                                                                                                                  
## Testing                                                                                                                                      
                                                                                                                                                  
  - **OS:** macOS 15 Sequoia (Darwin 24.6.0)                                                                                                      
  - Ran app overnight with 50+ network devices discovered                                                                                         
  - Socket count remained stable instead of growing unboundedly                                                                                   
  - USB-connected devices still establish tunnels correctly

### DIFF
--- a/ios/tunnel/tunnel_api.go
+++ b/ios/tunnel/tunnel_api.go
@@ -311,12 +311,23 @@ func RefreshTunnelForDevice(udid string, tunnelInfoHost string, tunnelInfoPort i
 
 // TunnelManager starts tunnels for devices when needed (if no tunnel is running yet) and stores the information
 // how those tunnels are reachable (address and remote service discovery port)
+// failedDevice tracks a device whose tunnel failed to start, so retries can be
+// backed off instead of attempted every UpdateTunnels cycle (each attempt opens
+// a usbmux socket, so hammering a device that always fails leaks sockets).
+type failedDevice struct {
+	lastAttempt time.Time
+	failCount   int
+}
+
 type TunnelManager struct {
-	ts                   tunnelStarter
-	dl                   deviceLister
-	pm                   PairRecordManager
-	mux                  sync.Mutex
-	tunnels              map[string]Tunnel
+	ts      tunnelStarter
+	dl      deviceLister
+	pm      PairRecordManager
+	mux     sync.Mutex
+	tunnels map[string]Tunnel
+	// failedDevices tracks devices whose tunnel start failed (keyed by udid) so
+	// UpdateTunnels can back off before retrying them.
+	failedDevices        map[string]failedDevice
 	startTunnelTimeout   time.Duration
 	firstUpdateCompleted bool
 	userspaceTUN         bool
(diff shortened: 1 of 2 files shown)


danielpaulus:
Thanks for the thorough writeup and the overnight `lsof` evidence, @vbragaru — this is a real leak and a well-reasoned fix. 🙏 `UpdateTunnels` opening a usbmux socket via `GetProductVersion` for every untracked device every cycle, with network devices that can never tunnel never being tracked, is exactly the accumulation you saw.

Since you opened this it had gone stale (`CONFLICTING`), and `main` migrated from logrus to our internal `golog` seam, so I took you up on "allow edits from maintainers" and rebased it onto current `main`:

- Ported both behaviours — skip `ConnectionType == "Network"`, and exponential backoff for failed devices (clear on success, prune on disconnect).
- Converted logging to `golog` with `module`/`udid` attrs.
- Pulled the skip decision and backoff schedule into pure functions (`shouldSkipDevice`, `failedDeviceBackoff`) and **added a test suite** using a stub device lister: network-skip, backoff hold-off, disconnect pruning, and failure recording. Nudged the first backoff to 30s to match your description (the original `1<<failCount` started at 60s).
- Built `currentUDIDs` before the per-device `udidFilter` check so the stale-entry pruning stays correct in per-device-agent mode.

Your authorship is preserved. gofmt/vet/`go test -race` all clean. Kicking off the real-device suite now.

/test-devices

PR #690, opened 2026-03-03
fix(windows): stop event loop on device disconnect

lizhizhuanshu:
Fix Windows event goroutine to exit when device.Events channel is closed on disconnect.

### DIFF
--- a/ios/tunnel/tunnel_windows.go
+++ b/ios/tunnel/tunnel_windows.go
@@ -50,10 +50,13 @@ func initTUNwrapper(device Device) *tunWrapper {
 	t.buffer = make([][]byte, 1)
 	t.buffer[0] = make([]byte, mtu)
 	go func() {
-		for {
-			e := <-device.Events()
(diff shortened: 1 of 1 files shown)


danielpaulus:
Thanks for catching this, @lizhizhuanshu! 🙏 You're right — `for { e := <-device.Events() }` busy-spins forever once the device disconnects and the channel is closed (a closed-channel receive returns the zero value immediately), pegging a core and leaking the goroutine. Ranging over the channel is the correct fix.

Since you opened this, the file was renamed (`tun_windows_wrapper.go` → `tunnel_windows.go`) and the package migrated from logrus to our internal `golog` seam, so the PR had gone stale/conflicting. I took you up on "allow edits from maintainers" and rebased your fix onto current `main`, applying the same change in the new location and log style (and a line noting the channel closed). Your authorship is preserved. Cross-compiles + vets clean for Windows. Merging once CI's green — thanks again! 🎉

PR #692, opened 2026-03-03
Add and remove wifi connection profiles

gbalduzzi:
Added support for a new command `wifi`, that installs a new profile to automatically connect to a given wifi network.

- When used on a supervised device and during the initial device setup, the profile installation does not require any input on the device
- Otherwise, the profile must be confirmed and activated manually on the device


danielpaulus (approved):
Nice command! Thanks!

PR #693, opened 2026-03-03
Added shutdown support

gbalduzzi:


danielpaulus (approved):
Thanks for adding this. I made a few changes and I am merging it now.

PR #694, opened 2026-03-03
Replace broadcast/subscribe with direct callbacks interface for acces…

buxomant:
## Summary                                                                                                                                                                                                 
                                                                                                                                                                                                             
  Replace the broadcast/subscribe notification pattern with a direct callbacks interface                                                                                                                     
  for accessibility inspector events.                                                                                                                                                                        
                                                                                                                                                                                                             
  ## Changes                                                                                                                                                                                                 
                                                                                                                                                                                                             
  - Add `AccessibilityInspectorCallbacks` interface with `HostAppStateChanged` and
    `HostInspectorNotificationReceived` methods
  - Replace `subscribers`/`sync.RWMutex` in `ControlInterface` with a single `callbacks` field
  - Remove `broadcast()` and `Subscribe()` methods
  - Simplify `Close()` — no longer needs to lock mutex or close subscriber channels
  - Both notification goroutines now use `ReceiveMethodCallWithTimeout` with a `context.Context`
    for clean shutdown, instead of blocking forever on `ReceiveMethodCall`
  - `hostAppStateChanged` now delivers parsed notifications to the caller
    (previously it only logged them)
  - `New()` accepts `context.Context` and `AccessibilityInspectorCallbacks` instead of a timeout

  ## Migration

  ```go
  // Before
  control, err := accessibility.New(device, 500*time.Millisecond)
  events, unsubscribe := control.Subscribe()
  defer unsubscribe()

  // After
  control, err := accessibility.New(device, ctx, myCallbacks)
  // notifications arrive via callback methods; cancel ctx to stop goroutines

### DIFF
--- a/main.go
+++ b/main.go
@@ -2176,14 +2176,9 @@ func timeFormat(device ios.DeviceEntry, operation string, force bool) {
 	}
 }
 
-type noopAxCallbacks struct{}
-
-func (noopAxCallbacks) HostAppStateChanged(accessibility.Notification)               {}
-func (noopAxCallbacks) HostInspectorNotificationReceived(accessibility.Notification) {}
-
 func startAx(device ios.DeviceEntry, arguments docopt.Opts) {
 	go func() {
-		conn, err := accessibility.New(device, context.Background(), noopAxCallbacks{})
+		conn, err := accessibility.NewWithoutEventChangeListeners(device)
 		exitIfError("failed starting ax", err)
 
 		conn.SwitchToDevice()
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -2176,9 +2176,14 @@ func timeFormat(device ios.DeviceEntry, operation string, force bool) {
 	}
 }
 
+type noopAxCallbacks struct{}
```
dmissmann: you could also move that into the accessibility package and use that in case `nil` is passed as the interface value.

PR #699, opened 2026-03-19
Add debugviewhierarchy: capture iOS view hierarchy via pure GDB RSP

### LINKED ISSUE #685: iOS View Hierarchy
XCode has a "Debug View Hierarchy"

It can be implemented via go-ios.

It mainly does a bunch of confusing madness using the debug proxy.

It also requires sending a RSP "DebugHierarchyPropertyAction" action.

It responds with gzip compressed hierarchy.

Good luck all.

aluedeke:
## Summary

Implements Xcode's "Debug View Hierarchy" feature for iOS 17+ devices using only Go — no LLDB, Python, or Xcode required.

- Connects to `com.apple.internal.dt.remote.debugproxy` via tunnel
- Attaches to the target process via GDB RSP
- Resolves `dlopen`/`objc_msgSend` addresses by parsing Mach-O export tries from process memory
- Injects `DebugHierarchyFoundation` + `libViewDebuggerSupport` via `dlopen()`
- Chains ObjC calls to execute `[[DBGTargetHub sharedHub] performRequestWithRequestInBase64:]`
- Reads gzipped JSON from process memory, decompresses, outputs to stdout

~1.5s per capture. Returns full view hierarchy with class names, frames, bounds, positions.

```
ios debugviewhierarchy <pid-or-bundle-id> | jq .
```

New package `ios/debuggertools` (3 files, ~640 lines):
- `viewhierarchy.go` — capture logic, reads like ObjC pseudocode
- `objc_runtime.go` — ObjC runtime abstraction (`Call`, `ClassCall`, `Dlopen`) over GDB RSP
- `macho.go` — Mach-O export trie parser for symbol resolution

Closes #685

## Test plan

- [ ] `ios debugviewhierarchy <pid>` outputs valid JSON with view hierarchy
- [ ] `ios debugviewhierarchy <bundle-id>` resolves PID and captures
- [ ] Verify with different app types (SwiftUI, UIKit)
- [ ] Test on different iOS versions (17.x, 18.x)
- [ ] Process resumes normally after capture (register state restored)


restapi/api/middleware.go
```diff
@@ -47,7 +49,40 @@ func DeviceMiddleware() gin.HandlerFunc {
...
+				log.WithField("udid", device.Properties.SerialNumber).
+					WithField("address", remotedInfo.Address).
+					Info("found existing remoted tunnel")
+				device, err = deviceWithRsdProvider(device, udid, remotedInfo.Address, remotedInfo.RsdPort)
+				if err != nil {
+					c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+					return
+				}
+			} else {
+				// Fall back to on-demand userspace lockdown tunnel for iOS 17.4+ (no sudo required)
```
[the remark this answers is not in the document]
nanoscopic: The difference is that below 17.4 you can't use the tcp tunnel and have to use the quic tunnel. You should know this.

main.go
```diff
@@ -907,6 +914,46 @@ The commands work as following:
...
+		if debugProxyPort == 0 {
+			exitIfError("debug proxy not found", fmt.Errorf("service com.apple.internal.dt.remote.debugproxy not in RSD"))
+		}
+
+		conn, connErr := ios.ConnectTUNDevice(device.Address, debugProxyPort, device)
+		exitIfError("connect to debug proxy", connErr)
+
+		gdb := debugserver.NewGDBServer(conn)
+		gdb.Request("QStartNoAckMode")
+		gdb.Request("qSupported:xmlRegisters=i386,arm,mips,arc")
```
dmissmann, line 940: `i386,arm,mips,arc` seems strange, none of those architectures are used (afaik `arm` refers to the 32-bit arm CPUs)

ios/debugserver/gdbserver.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (g *GDBServer) formatPacket(pck string) string {
+	if g.noAck {
+		return "$" + pck + "#" + g.chksum(pck)
+	}
+	return "+$" + pck + "#" + g.chksum(pck)
+}
+
+// SetNoAckMode should be called after QStartNoAckMode succeeds.
+func (g *GDBServer) SetNoAckMode() {
```
dmissmann, line 36: That's unused, but `QStartNoAckMode` is being sent. And if calling this is a requirement after sending `QStartNoAckMode`, this should be done inside a method.

ios/debugserver/gdbserver.go
```diff
@@ -0,0 +1,99 @@
...
+	for {
+		_, err := io.ReadFull(g.rw, oneByte)
+		if err != nil {
+			return "", fmt.Errorf("recv: waiting for '$': %w", err)
+		}
+		if oneByte[0] == '$' {
+			break
+		}
+		// Skip +, -, and any other non-$ bytes
+	}
```
dmissmann, line 61: maybe it's worth using a buffered reader here. We always trigger a Read call on a `net.Conn` for reading single bytes.

ios/debuggertools/viewhierarchy.go
```diff
@@ -0,0 +1,152 @@
...
+	request, err := rt.ClassCall("NSString", "stringWithUTF8String:", rt.CString(b64))
+	if err != nil {
+		rt.cleanup()
+		return nil, fmt.Errorf("[NSString stringWithUTF8String:]: %w", err)
+	}
+
+	return &ViewHierarchySession{rt: rt, hub: hub, request: request}, nil
+}
+
+// Dump captures the current view hierarchy and returns decompressed JSON.
```
dmissmann, line 77: do we know the structure of this JSON data?

main.go
```diff
@@ -909,6 +916,46 @@ The commands work as following:
...
+		gdb.Request("QEnableErrorStrings")
+		gdb.Request("QThreadSuffixSupported")
+		gdb.Request("QListThreadsInStopReply")
```
dmissmann, line 943: are all those requests done here necessary to make it work? If so can we move that to `SetupViewHierarchy`?

dmissmann (commented):
don't know really anything about the internals of this. using this seems (almost 🙂 ) straight forward since it doesn't involve a lot of exported functions and if this was tested it should be ok

PR #706, opened 2026-04-14
Add os_trace_relay service for structured syslog streaming

aluedeke:
## Summary

Adds support for `com.apple.os_trace_relay`, the modern structured syslog service available on iOS 10+. Unlike the legacy `syslog_relay`, it provides:

- **Device-side filtering** by PID and log level (`MessageFilter` bitmask) — reduces USB traffic on noisy devices
- **Structured binary entries** with process name, log level, subsystem/category labels, thread ID, image name/offset — no regex parsing
- **Optional client-side filters** for subsystem substring match, message include/exclude

Binary parser follows the `ostrace_packet_header_t` struct from [libimobiledevice/ostrace.h](https://github.com/libimobiledevice/libimobiledevice/blob/master/include/libimobiledevice/ostrace.h). This fixes a few field-size errors present in [pymobiledevice3's parser](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/services/os_trace.py) (treats `time_sec` as uint64, `message_len` as uint32, and `subsystem_len`/`category_len` as uint16).

Follows the same dual-transport pattern as `syslog` — usbmuxd on pre-iOS-17, CoreDevice tunnel (`com.apple.os_trace_relay.shim.remote`) on iOS 17+.

## New CLI

```
ios ostrace [--pid=<pid>] [--process=<name>] [--level=<levels>]
            [--subsystem=<sub>] [--match=<str>] [--exclude=<str>]
```

Device-side filters:
- `--pid` — stream only logs from this PID
- `--process` — resolve process name → PID via instruments, then filter device-side
- `--level` — comma-separated levels: `notice,info,debug,useraction,error,fault`

Client-side filters (applied locally):
- `--subsystem`, `--match`, `--exclude`

## New REST endpoint

`GET /device/:udid/ostrace?pid=<pid>&level=<levels>&subsystem=<sub>&match=<str>&exclude=<str>`

Streams newline-delimited JSON entries. Properly releases the device connection via `defer conn.Close()`.

## Test plan

- [x] Unit tests for binary parser, filter parsing, level strings (`go test ./ios/ostrace/`)
- [x] `go build ./...` passes
- [x] `go vet ./...` clean
- [x] Verified against real iOS 18.7.1 device — streams structured entries with pid/level/subsystem labels
- [x] Verified `--pid` filter works device-side (compared entry count with and without filter)
- [x] Verified `--level=error,fault` reduces traffic (hardcoded `0xFFFF` would send everything)

## Notes

- The `--process` option depends on the Instruments service which requires the Developer Disk Image. All other paths work without DDI.
- Adds a reasonable amount of documentation distinguishing device-side vs client-side filtering in `ios help` output.

### DIFF
--- a/ios/ostrace/ostrace.go
+++ b/ios/ostrace/ostrace.go
@@ -43,16 +43,29 @@ const (
 
 // StreamFlags gates SEVERITY for record types that carry a level
 // (ActivityTransition and LogMessage). Default, Error, and Fault always
-// emit regardless of flags; Info and Debug require StreamFlagsDebug (0x20).
+// emit unconditionally; Info and Debug require StreamFlagsDebug (0x20).
 //
-// Empirically verified on iOS 18.7.1: bit 0x20 is the ONLY bit that unlocks
-// Info or Debug entries (brute-forced all 16 bits individually and several
-// combinations). When 0x20 is set, Info AND Debug always flow together —
-// no bit separates them. The 0x100 bit from reverse-engineering notes of
-// LoggingSupport/diagnosticd is a no-op on iOS 18. Clients wanting only
-// Info or only Debug must filter client-side.
+// Info/Debug cannot be separated at the wire level on stock iOS 18.
+// The kernel's /dev/oslog_stream decides at the source whether Info/Debug
+// records exist — controlled by the arm64 commpage at 0xfffffc104:
+//
+//	bit 0 = preserve Default, bit 1 = preserve Info, bit 3 = streaming active
+//
+// When diagnosticd processes StreamFlags it calls
+// host_set_atm_diagnostic_flag(mode & ~3), stripping bits 0 and 1 — so the
+// StreamFlags value alone can never turn on Info preservation. Setting 0x20
+// triggers mode=0xb, which also invokes _os_trace_set_mode(0xb) and flips
+// the per-process preservation bits for the streaming session. That's what
+// enables Info and Debug to flow together. 0x100 (mode=9) skips that call,
+// which is why it's a no-op on stock iOS despite passing diagnosticd's
+// userspace gate. Only `log config --mode level:info` (root-entitled) can
+// set commpage bit 1 system-wide.
+//
+// Verified on iOS 18.7.1 by a full 16-bit single-bit sweep plus several
+// multi-bit combos: no bit or combo separates Info from Debug. Clients
+// wanting only one must filter client-side.
 const (
-	StreamFlagsDebug uint32 = 0x20 // enables Info and Debug entries
+	StreamFlagsDebug uint32 = 0x20 // triggers mode=0xb → enables Info+Debug together
 	StreamFlagsAll   uint32 = StreamFlagsDebug
 )
 
--- a/ios/ostrace/ostrace.go
+++ b/ios/ostrace/ostrace.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"slices"
 	"strings"
 	"time"
 
@@ -285,7 +286,7 @@ func (c *Connection) startActivity(pid int, messageFilter uint16, streamFlags ui
 	}
 
 	// Reverse the bytes and interpret as big-endian integer
-	reverseBytes(lengthBytes)
+	slices.Reverse(lengthBytes)
 	var plistLength uint64
 	for _, b := range lengthBytes {
 		plistLength = (plistLength << 8) | uint64(b)
@@ -396,6 +397,9 @@ func parseEntry(data []byte) (LogEntry, error) {
 	}
 
 	pid := binary.LittleEndian.Uint32(data[9:13])
+	if pid > 999999 {
+		return LogEntry{}, fmt.Errorf("ostrace: pid %d exceeds sanity limit, possibly corrupted stream", pid)
+	}
 	procpathLen := binary.LittleEndian.Uint16(data[37:39])
 	timeSec := binary.LittleEndian.Uint64(data[55:63])
 	timeUsec := binary.LittleEndian.Uint32(data[63:67])
@@ -475,9 +479,3 @@ func cstring(data []byte) string {
 	return string(data)
 }
 
-// reverseBytes reverses a byte slice in place.
-func reverseBytes(b []byte) {
-	for i, j := 0, len(b)-1; i < j; i, j = i+1, j-1 {
-		b[i], b[j] = b[j], b[i]
-	}
-}
--- a/ios/ostrace/ostrace.go
+++ b/ios/ostrace/ostrace.go
@@ -28,47 +28,15 @@ const (
 	LogLevelFault      LogLevel = 0x11
 )
 
-// MessageFilterAll includes all log levels.
-const MessageFilterAll uint16 = 0xFFFF
+// messageFilterAll is the MessageFilter value used in StartActivity requests.
+// The individual bit semantics of this field are undocumented; pymobiledevice3
+// and libimobiledevice both hardcode 0xFFFF and we mirror that.
+const messageFilterAll uint16 = 0xFFFF
 
-// ParseMessageFilter parses a comma-separated list of level names into a MessageFilter bitmask.
-// Valid names: notice, info, debug, useraction, error, fault (case-insensitive).
-// Returns MessageFilterAll if the input is empty.
-func ParseMessageFilter(levels string) (uint16, error) {
-	if levels == "" {
-		return MessageFilterAll, nil
-	}
-	var filter uint16
-	for _, name := range splitAndTrim(levels) {
-		bit, ok := levelNameToBit[strings.ToLower(name)]
-		if !ok {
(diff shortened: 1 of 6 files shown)


ios/ostrace/ostrace.go
```diff
@@ -0,0 +1,387 @@
...
+type Connection struct {
+	closer io.Closer
+	reader io.Reader
+	writer io.Writer
+}
+
+// New creates a new os_trace_relay connection, sends the StartActivity request
+// with the given pid filter and messageFilter bitmask, and performs the handshake.
+// Use pid=-1 for all processes and messageFilter=0xFFFF for all log levels.
+func New(device ios.DeviceEntry, pid int, messageFilter uint16) (*Connection, error) {
```
dmissmann: I assume `messageFilter` here is related to the logging levels, or which values other than `0xFFFF` are possible? If not, what's the difference between the filter and a level?

ios/ostrace/ostrace.go
```diff
@@ -0,0 +1,387 @@
...
+
+// startActivity sends the StartActivity plist request and reads the handshake response.
+func (c *Connection) startActivity(pid int, messageFilter uint16) error {
+	codec := ios.NewPlistCodecReadWriter(c.reader, c.writer)
+
+	request := map[string]interface{}{
+		"Request":       "StartActivity",
+		"MessageFilter": messageFilter,
+		"Pid":           pid,
+		"StreamFlags":   60,
```
dmissmann: Do you know what this is?

main.go
```diff
@@ -2731,6 +2768,51 @@ func runSyslog(device ios.DeviceEntry, parse bool) {
...
+	go func() {
+		for {
+			entry, err := conn.ReadEntry()
+			if err != nil {
+				exitIfError("failed reading os_trace entry", err)
+			}
+			if !clientFilter.Matches(entry) {
+				continue
+			}
+			fmt.Println(convertToJSONString(entry))
```
dmissmann: I'm fine with that implementation, but what do you think about making reading the messages part of the filter so that we don't have to call `Matches` ourselves and just get the entries that match?

dmissmann (commented):
Maybe a similar warning as in the Console app should be added somewhere (it says `Streaming log messages will impact system performance.`) as this adds a lot of CPU load

aluedeke:
Thanks for the review @dmissmann! Addressed all four in 3504fa4:

**1. `messageFilter` semantics (line 150)** — I tried to document what the bits mean and in the process discovered my bit-mapping was speculative. Empirical testing showed e.g. `0x01` returned Info-level entries (not Notice as my mapping assumed) and `0x10` returned nothing. Neither pymobiledevice3 nor libimobiledevice document the bits either — both just hardcode `0xFFFF`. So I removed the parser and hardcoded `0xFFFF` with a comment explaining it's undocumented. The `--level` CLI flag stays but now filters **client-side** using the parsed `LogLevel` values (which ARE documented in libimobiledevice's public header). Same UX, honest about what happens.

**2. `StreamFlags: 60` (line 185)** — Added a comment noting `0x3C` is mirrored from pymobiledevice3/libimobiledevice and its bits are undocumented. Removed from any configurable surface, kept internal only. If we ever figure out what it means we can expose it later.

**3. Filter as part of reading (line 2808)** — Added `Connection.ReadFilteredEntry(filter ClientFilter)` that loops internally until a match is found. Both the CLI and REST endpoint now use it instead of calling `Matches` themselves.

**4. Performance warning** — Added `log.Warn("Streaming log messages places significant CPU load on the device.")` at CLI startup, mirroring Console.app's behavior. Also mentioned in the help text.

Verified against a real iOS 18.7.1 device: `--level=error` filters correctly (30 entries, all "Error" level in a 3s window), warning shows at startup.

main.go
```diff
@@ -2731,6 +2770,49 @@ func runSyslog(device ios.DeviceEntry, parse bool) {
 	<-c
 }
 
+func runOsTrace(device ios.DeviceEntry, pid int, processName string, clientFilter ostrace.ClientFilter) {
+	log.Debug("Run OsTrace.")
+	log.Warn("Streaming log messages places significant CPU load on the device.")
```
gmegidish, line 2775: should we have this? it's part of documentation I imagine, not a warning of the user doing something wrong. wdyt?

main.go
```diff
@@ -2731,6 +2770,49 @@ func runSyslog(device ios.DeviceEntry, parse bool) {
...
+	if processName != "" && pid == -1 {
+		service, err := instruments.NewDeviceInfoService(device)
+		exitIfError("failed opening deviceInfoService for resolving process name", err)
+		procs, err := service.ProcessList()
+		service.Close()
+		exitIfError("failed getting process list", err)
+
+		found := false
+		for _, p := range procs {
+			if p.Name == processName || p.RealAppName == processName {
```
gmegidish, line 2786: this code is the same for memlimitoff. may we should refactor both to have getProcessId?

ios/ostrace/ostrace.go
```diff
@@ -0,0 +1,414 @@
...
+type ClientFilter struct {
+	Levels    []LogLevel // If non-empty, only entries whose Level is in this set pass
+	Subsystem string     // Only show entries where label subsystem contains this string
+	Match     string     // Only show entries where message contains this string
+	Exclude   string     // Hide entries where message contains this string
+}
+
+// Matches returns true if the entry passes all client-side filters.
+func (f ClientFilter) Matches(entry LogEntry) bool {
+	if len(f.Levels) > 0 {
```
gmegidish, line 102: wouldn't a simple >= level be enough? do we need an array? does iOS have more levels that the 1,2,3,0x10,0x11 written above, so we don't want to show 4,5,6,7,8,9,0xa,0xb,0xc,0xd,0xe,0xf? wdyt?

main.go
```diff
@@ -2731,6 +2770,49 @@ func runSyslog(device ios.DeviceEntry, parse bool) {
...
+
+	conn, err := ostrace.New(device, pid)
+	exitIfError("os_trace connection failed", err)
+	defer conn.Close()
+
+	go func() {
+		for {
+			entry, err := conn.ReadFilteredEntry(clientFilter)
+			if err != nil {
+				exitIfError("failed reading os_trace entry", err)
```
gmegidish, line 2806: exitIfError will call Fatalf & os.exit(). defer conn.Close() not gonna happen. are we okay with this?

ios/ostrace/ostrace.go
```diff
@@ -0,0 +1,414 @@
...
+		return ""
+	}
+	if data[len(data)-1] == 0 {
+		return string(data[:len(data)-1])
+	}
+	return string(data)
+}
+
+// reverseBytes reverses a byte slice in place.
+func reverseBytes(b []byte) {
```
gmegidish, line 479: slices.Reverse(b) ?

ios/ostrace/ostrace.go
```diff
@@ -0,0 +1,414 @@
...
+//	~       var   cstring    imagepath (imagepathLen bytes)
+//	~       var   cstring    message (messageLen bytes)
+//	~       var   cstring    subsystem (subsystemLen bytes, optional)
+//	~       var   cstring    category (categoryLen bytes, optional)
+func parseEntry(data []byte) (LogEntry, error) {
+	if len(data) < 129 {
+		return LogEntry{}, fmt.Errorf("ostrace: entry too short: %d bytes", len(data))
+	}
+
+	pid := binary.LittleEndian.Uint32(data[9:13])
```
gmegidish, line 398: should we add a check if see if pid is a valid value? like, if it's beyond 1,000,000 then it means we have a corrupted stream and should abort? wdyt?

gmegidish (approved):

ios/ostrace/ostrace.go
aluedeke: Good question! A simple `>=` doesn't work here because the level values aren't contiguous — they go `0x00, 0x01, 0x02, 0x03` then jump to `0x10, 0x11`:

```
Default    = 0x00
Info       = 0x01
Debug      = 0x02
UserAction = 0x03
Error      = 0x10
Fault      = 0x11
```

So `>= Error (0x10)` would correctly include Fault (0x11) but miss everything below 0x10. And `>= Info (0x01)` would include UserAction (0x03) and all the unknown values 0x04–0x0F. The array approach lets users pick arbitrary combinations like `error,fault` or `default,error` without assumptions about ordering or gaps. It's a small array (5 entries max) so the O(n) scan per entry is negligible vs. the I/O cost.

aluedeke:
Thanks @gmegidish for the thorough review! All addressed in 1e17e44:

1. **Warning removed** — kept only in help text, not log output
2. **ProcessByName** — extracted into `DeviceInfoService.ProcessByName()`, used by both `ostrace` and `memlimitoff` now
3. **Levels array** — replied inline with explanation (level values aren't contiguous: 0x00-0x03 then 0x10-0x11, so `>=` doesn't work)
4. **defer/exitIfError** — goroutine now signals via channel, `conn.Close()` runs before `exitIfError`
5. **reverseBytes** → `slices.Reverse` ✓
6. **PID sanity check** — added `pid > 999999` guard in `parseEntry`

PR #714, opened 2026-05-06
Add wallpaper and icon-layout commands

aluedeke:
## Summary

Adds four CLI commands for inspecting and modifying iOS home-screen state on supervised devices, reverse-engineered from Apple Configurator's `cfgutil set-wallpaper` and the `com.apple.springboardservices` RPC.

- **`ios set-wallpaper <image> [--screen=lock|home|both] --p12file=... [--password=...]`** — sets the device wallpaper. Sends the MDM `Settings` command with item `Wallpaper` over the post-Escalate `com.apple.mobile.MCInstall` channel, exactly matching cfgutil's wire format (verified by DTrace).
- **`ios get-wallpaper [--output=wallpaper.png]`** — pulls the home-screen wallpaper PNG via `springboardservices`. No supervision required.
- **`ios get-icon-layout [--output=…]`** / **`ios set-icon-layout <file>`** — round-trippable JSON capture/restore of the home-screen layout (`getIconState` / `setIconState`).

The supervised auth handshake (Escalate / EscalateResponse / ProceedWithKeybagMigration) reuses the existing `mcinstall.EscalateWithCertAndKey` path; no new crypto is added.

## Behavior caveats (documented in CLI help and Go doc-comments)

- **iOS 16+ unified Lock + Home wallpaper:** the device applies the image to both screens regardless of the `Where` field. Apple's own cfgutil exhibits the same behavior. The `--screen` flag is preserved for older iOS / forward-compat.
- **`get-wallpaper` may EOF on iOS 18:** known iOS-side regression also seen in pymobiledevice3 ([doronz88/pymobiledevice3#1450](https://github.com/doronz88/pymobiledevice3/issues/1450)). Code is correct.
- **`set-icon-layout` constraint:** iOS requires every installed app to occupy a slot. Per cfgutil docs: *"unexpected behavior may occur if the given layout does not contain every icon on the device".* Missing apps are re-paginated, not hidden.
- **Edit-Pages hidden-page bit (iOS 14+) is not exposed** by springboardservices, so a fetched layout will not include hidden pages, and pushing a layout cannot reproduce a hidden page.

## Test plan

- [x] `go build ./...` clean (only my touched packages; existing untracked `ios/resign/` WIP excluded)
- [x] `go vet ./ios/mcinstall ./ios/springboard ./` clean
- [x] `set-wallpaper` end-to-end on a real supervised iPhone SE (iOS 18.7.1) — wallpaper changes verified visually
- [x] `--screen=lock|home|both` all return `"ok"`; behavior matches cfgutil exactly (both screens get set on iOS 16+, unavoidable)
- [x] `get-icon-layout` → `set-icon-layout` round-trip on real device returns `"ok"`, layout preserved
- [x] Negative tests: invalid `--screen`, missing `--p12file`, `P12_PASSWORD` env-var fallback, all behave correctly
- [ ] `get-wallpaper` on iOS < 18 (no test device available)

### DIFF
--- a/ios/mcinstall/mcinstall.go
+++ b/ios/mcinstall/mcinstall.go
@@ -332,6 +332,81 @@ func (mcInstallConn *Connection) AddProfileSupervised(profileFileBytes []byte, p
 	return mcInstallConn.addProfile(profileFileBytes, "InstallProfileSilent")
 }
 
+// WallpaperScreen identifies which screen(s) to apply a wallpaper to. Values
+// match the MDM "Wallpaper" Settings command's "Where" field.
+type WallpaperScreen int
+
+const (
+	WallpaperLockScreen WallpaperScreen = 1
+	WallpaperHomeScreen WallpaperScreen = 2
+	WallpaperBoth       WallpaperScreen = 3
+)
+
+// ParseWallpaperScreen maps "lock", "home" or "both" to the corresponding
+// WallpaperScreen value. Matches the cfgutil --screen flag.
+func ParseWallpaperScreen(s string) (WallpaperScreen, error) {
+	switch s {
+	case "lock":
+		return WallpaperLockScreen, nil
+	case "home":
+		return WallpaperHomeScreen, nil
+	case "both":
+		return WallpaperBoth, nil
+	}
+	return 0, fmt.Errorf("invalid screen %q (expected lock|home|both)", s)
+}
+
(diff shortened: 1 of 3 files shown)


danielpaulus (approved):
Reviewed: builds/vets/gofmt clean merged with main; all mcinstall/springboard symbols present; logic sound; tested on a real supervised device by the author. Minor non-blocking note: SetIconLayout doesn't read a response — fine for now.

PR #718, opened 2026-05-10
ostrace: fix --nojson flag and add ANSI color-coded output for plain text mode

zt-luo:
- Fix `--nojson` flag: use structured plain-text output via `formatEntryPlain` instead of raw JSON dump
- Add human-readable plain-text output with ANSI color coding:
  - Info → cyan, Debug → gray, Error → red, Fault → bold red
  - Subsystem/category label dimmed in gray
  - Timestamp in RFC3339 format
- Add `isTerminal()` check: color codes are only emitted when stdout is a TTY (clean output in pipes/files)
- Format: `[timestamp] PID:##### <Level> [subsystem:category] message`


danielpaulus:
Thanks @zt-luo! The real-device CI caught a cross-platform build break: `isTerminal()` uses `unix.IoctlGetTermios(fd, unix.TCGETS)`, but `golang.org/x/sys/unix` is Unix-only (breaks Windows) and `TCGETS` is Linux-only (breaks macOS — it's `TIOCGETA` there).

I'll push a small commit on top swapping it for the portable `golang.org/x/term.IsTerminal(int(fd))` (already in the module graph), re-run the device tests, and merge with a merge commit so your authorship stays intact. 🙏

PR #719, opened 2026-05-10
ostrace: add --follow flag for persistent log tracking across process restarts

zt-luo:
- Add `--follow` flag to `ostrace` command for persistent log tracking
- When enabled, the client auto-reconnects if the process exits or restarts
- Polls `ProcessList()` every 3 seconds to detect process exit via `isProcessAlive()`
- Uses `context.WithCancel` for clean goroutine lifecycle — monitor goroutine is cancelled on reconnect/error to avoid leaks
- Non-blocking send on reconnect channel prevents potential deadlock
- With explicit `--pid` (no `--process`), exits gracefully when the PID terminates instead of looping
- With `--process=<name>`, auto-resolves the new PID on process restart
- `isProcessAlive()` logs warnings on transient connection errors for debuggability
- `sleepOrCancel()` replaces `time.Sleep` for instant Ctrl+C response


danielpaulus (approved):
Thanks! Cool stuff!

PR #720, opened 2026-05-10
imagemounter: fetch DDI from GitHub and improve mount reliability

zt-luo:
- Switch iOS 17+ DDI source from deviceboxhq.com to github.com/doronz88/DeveloperDiskImage (build 17E5179g), resolving identity-not-found failures on newer chips (A19 Pro)
- Add QueryPersonalizationManifest fast path: try existing device-side manifest before requesting Apple TSS, and reconnect the service on failure since the socket is closed by the device. This skips nonce + TSS on re-mounts when the manifest is still valid
- Replace hardcoded Manifest struct with dynamic map entries and iterate all Trusted entries when building the TSS request
- Update TSS @VersionInfo and add @UUID field
- Add ProductBuildVersion-based DDI cache validation and stream DMG/trustcache directly to disk during download


danielpaulus:
Awesome stuff! I would rather keep the DDI on deviceboxhq.com because it has a fast global CDN, so I made a few changes. Thanks a ton for this improvement!

danielpaulus (approved):
Awesome, thanks a ton!!

PR #736, opened 2026-06-03
refactor(log): migrate logrus -> slog (ios/golog) and remove logrus

danielpaulus:
Migrates all of `main.go` + `ios/**` from logrus to `log/slog`, and **removes logrus from the root module**. Done largely via a per-package agent migration; every package was build/vet/gofmt-verified.

## The injection seam (your config knob)
- **`ios/golog`** — go-ios's internal logging package: thin slog wrapper (`Trace/Debug/Info/Warn/Error/Fatal`, `Enabled`, `LevelTrace`). Defaults to `slog.Default()`.
- **`ios.SetLogger(l *slog.Logger)`** — the public opt-in. Routes **only go-ios's** logs to your logger; it does **not** touch the process-global `slog.Default()`, so an importer's own logging is untouched. Call nothing → standard slog behavior, unchanged.

```go
ios.SetLogger(slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{Level: ios.LevelTrace})))
```

## Conventions applied
- printf calls restructured into slog key/value attrs (`Debugf("flush: %v", re)` → `golog.Debug("flush response", "response", re)`). `fmt.Sprintf` kept only as an *attr value* for verbs slog lacks (`%x`/`%T`/`%#v`).
- `WithFields/WithError` → attrs; `Fatal` → error+`os.Exit(1)`; level guards → `golog.Enabled(...)`.
- Custom **`LevelTrace`** (= `slog.LevelDebug-4`) preserves `--trace`.

## CLI behavior preserved (`main.go`)
Builds one slog handler from flags and installs it via `ios.SetLogger`: **JSON to stderr by default, text on `--nojson`, level from `--verbose`/`--trace`** (with a `TRACE` level label). Smoke-tested: default/`--nojson`/`--trace` all behave as before; command JSON output (stdout) unchanged.

## Public API
No exported `ios/**` signature changes **except** `ios/debugproxy.NewDtxDecoder` / `NewBinDumpOnly`: their param was already a logrus type (`*logrus.Entry`) so it could not survive logrus removal — now `*slog.Logger`. Both return an unexported type and have **no callers outside the package**, so practical impact is nil.

## Scope / out of scope
- Root module only. `restapi/`, `ncm/`, `cmd/` are **separate modules** and keep their own logging untouched.

## Verification
- `go build ./...`, `go test ./...`, `gofmt` all clean.
- `go vet ./...` only reports 3 pre-existing `debugproxy` copylocks warnings (value receivers embedding `sync.Mutex`; present on `main`, unrelated to this change).

Large diff (71 files) but mechanical and per-package verified — worth a skim of the `ios/golog` package, the `main.go` handler setup, and the debugproxy signature note.


ios/golog/golog.go
```diff
@@ -0,0 +1,54 @@
...
+
+// L returns the logger go-ios currently logs through.
+func L() *slog.Logger {
+	if l := logger.Load(); l != nil {
+		return l
+	}
+	return slog.Default()
+}
+
+func Trace(msg string, args ...any) { L().Log(context.Background(), LevelTrace, msg, args...) }
```
danielpaulus: With `AddSource` enabled, these wrappers will report `ios/golog/golog.go` as the source rather than the package that called `golog.Info`/`Debug`/etc. `slog.Logger.Log` and the level helpers capture the program counter at the wrapper callsite, so this conflicts with the new AGENTS guidance to rely on `AddSource` for call sites and makes source attribution misleading for every library log. Can we preserve the caller PC in the seam (for example via `runtime.Callers` + `slog.NewRecord`/`Handler.Handle`) or change the documented contract?

ios/golog/golog.go
```diff
@@ -0,0 +1,54 @@
...
+func Debug(msg string, args ...any) { L().Debug(msg, args...) }
+func Info(msg string, args ...any)  { L().Info(msg, args...) }
+func Warn(msg string, args ...any)  { L().Warn(msg, args...) }
+func Error(msg string, args ...any) { L().Error(msg, args...) }
+
+// Fatal logs at error level and exits the process with status 1, matching the
+// behavior of the logrus Fatal calls this replaces.
+func Fatal(msg string, args ...any) {
```
danielpaulus: This preserves the old logrus Fatal behavior, but now it becomes part of the new library logging seam. Long term I would avoid exposing/encouraging `golog.Fatal` in `ios/*`: library code exiting the process is surprising for embedders using `ios.SetLogger`. For this migration, can we at least keep `Fatal` usage constrained to CLI-like paths/tests, or add a follow-up to return errors from the active library callsites instead?

main.go
```diff
@@ -859,8 +904,9 @@ The commands work as following:
 
 	b, _ = arguments.Bool("dproxy")
 	if b {
-		log.SetFormatter(&log.TextFormatter{})
-		// log.SetLevel(log.DebugLevel)
+		// NOTE: previously this forced a logrus TextFormatter for the dproxy
```
danielpaulus, line 907: This drops the old dproxy-specific text formatter, so `ios dproxy` now uses the global default JSON handler unless users pass `--nojson`. That may be fine, but it is a user-visible behavior change outside the mechanical logrus->slog migration. If dproxy output is consumed by humans/scripts, I would either preserve the text override here or document/test the new default explicitly.

main.go
```diff
@@ -1506,7 +1552,7 @@ The commands work as following:
 			defer outputFile.Close()
 
 			// Download file using streaming to minimize memory usage
-			log.Infof("Downloading %s to %s...", remotePath, localPath)
+			slog.Info("downloading", "remote", remotePath, "local", localPath)
```
danielpaulus: Several user-facing CLI messages now become structured slog text records when `--nojson` is set, e.g. `msg=downloading remote=... local=...` instead of the previous formatted sentence. That is not necessarily wrong, but it is a CLI output compatibility change mixed into the logger migration. Worth doing a deliberate pass over the no-json paths and either keeping the old human strings where they are command output, or documenting that `--nojson` now means slog text output rather than the old logrus text shape.

danielpaulus (commented):
Overall I like the direction of this refactor. Moving the root module from logrus to stdlib `slog`, adding an internal `ios/golog` seam, and exposing `ios.SetLogger` is the right shape for a project that is both a CLI and an embeddable library. The dependency reduction and structured attrs (`module`, `udid`, ids/paths/pids) are real improvements.

The one thing I would fix before merging is the caller/source behavior in `golog`: with `AddSource`, the new seam reports the wrapper file instead of the actual callsite, which conflicts with the new logging guidance and will make diagnostics misleading.

The other comments are compatibility/design checks rather than hard blockers: `dproxy` losing its text formatter override, `--nojson` human output changing shape in some commands, and `golog.Fatal` remaining available in library code. Those may be acceptable if intentional, but I would make them explicit.

Verification from an isolated PR checkout:
- `go test ./...` passes.
- `go test ./... -run "^$"` passes.
- `go vet ./...` reports existing-looking debugproxy mutex-copy warnings, not obviously introduced by this slog migration.

ios/golog/golog.go
danielpaulus: Fixed in e97b63b — `golog` now captures the caller's PC (`runtime.Callers` + `slog.NewRecord` + `Handler.Handle`), so `AddSource` reports the real callsite instead of `golog.go`. Added a regression test (`ios/golog/golog_test.go`) asserting the resolved source is the caller, not the wrapper. Cost matches calling slog directly (one `runtime.Callers`, no extra allocation), so the AGENTS `AddSource` guidance now holds.

ios/golog/golog.go
danielpaulus: Done in 91e2373 — removed `golog.Fatal` entirely. The 4 callsites now return errors instead of exiting: `debugserver.Start` returns the listen error and the lldb result via a channel (no more `os.Exit(0)`/`Fatal`, so it's embeddable); the screenshot goroutine logs at error and stops its loop instead of killing the host; `zipconduit` `init()` panics on the impossible hardcoded-const decode (init can't return). `main.go`'s `ios debug` uses `exitIfError` to preserve the non-zero exit on failure.

main.go
danielpaulus: Intentionally left as-is: `ios dproxy` now follows the global handler — JSON by default, human-readable text with `--nojson`. We chose to accept the unified default rather than special-case a per-command formatter.

main.go
danielpaulus: Fixed in 5f32fdc — reverted the genuinely user-facing feedback lines (file-transfer progress, port forwarding, profile add/remove, pairing, `--follow` status, signal handling) back to readable sentences. Structured key/value attrs remain for internal diagnostics.

danielpaulus (commented):
Second pass on current head `07331f9e`:

The important fixes from the earlier review look good. `golog` now captures the caller PC explicitly and has `TestSourceReportsCallerNotWrapper`, so `AddSource` should report the real callsite. The library-level `golog.Fatal` concern is also addressed: the seam no longer exposes `Fatal`, debugserver returns errors, screenshot streaming stops its loop instead of exiting the host process, and the impossible zipconduit init failure is now a panic on a hardcoded constant.

I did not find a new blocking issue in this round. Remaining things I would still treat as compatibility checks rather than correctness blockers:
- `dproxy` still no longer forces text formatting, so default output is governed by the global JSON/`--nojson` choice.
- A few CLI messages are intentionally kept as formatted slog messages to preserve human text, while others are structured attrs. That is fine if deliberate, but it is worth keeping the distinction clear: command output compatibility vs diagnostic logs.

Verification:
- `go test ./...` passes.
- `go test ./... -run TestSourceReportsCallerNotWrapper -count=1` passes.
- `go vet ./...` still reports the existing debugproxy mutex-copy warnings; I did not see a new vet issue from this PR head.

ios/instruments/screenshot.go
```diff
@@ -100,10 +100,13 @@ func startScreenshotting(conn *ScreenshotService) {
 		start := time.Now()
 		pngBytes, err := conn.TakeScreenshot()
 		if err != nil {
-			log.Fatal("Screenshot failed", err)
+			// Stop the streaming loop instead of killing the host process; a
+			// screenshot failure must not take down a caller embedding go-ios.
+			golog.Error("screenshot failed, stopping screenshot loop", "module", logModule, "error", err)
+			return
```
danielpaulus, line 106: This is the one `log.Fatal` migration that looks behaviorally risky. `StartMJPEGStreamingServer` still blocks in `http.ListenAndServe` after this goroutine returns, and `mjpegHandler` waits on per-client channels that only receive frames from this loop. So after the first `TakeScreenshot` error, the command can stay alive serving clients that block forever instead of the old behavior where `log.Fatal` terminated the process with a failure. I agree library code should not call `os.Exit`, but this path needs an error propagation/shutdown mechanism, e.g. run the screenshot loop with an error channel/context and close the HTTP server when capture fails.

danielpaulus (commented):
Fatal-call audit, ignoring dproxy:

I reviewed the changed `log.Fatal`/fatal-like paths against `main`.

Looks behavior-preserving:
- CLI argument validation (`--orgname`, launch/kill/debug missing args, set-wallpaper `--p12file`) now goes through `logFatal`, which still logs and exits with status 1.
- `exitIfError` still logs and exits with status 1; only the logger implementation changed.
- `debugserver.Start` now returns listener/lldb errors, and the CLI wraps it with `exitIfError`, so CLI failure still exits non-zero while library callers no longer get killed. This is a good semantic improvement.
- `zipconduit` hardcoded hex failure changes from `log.Fatal` to `panic`; since this is an impossible programmer error during init, that is fine and arguably clearer.
- Removing the block-commented AFC `log.Fatalf` tests has no runtime effect.

One risky change is inline on `ios/instruments/screenshot.go`: the MJPEG screenshot loop now returns on capture failure, but the HTTP server keeps running and handlers can block waiting for frames. That should propagate failure/shutdown instead of silently leaving a dead stream server alive.

PR #737, opened 2026-06-03
Make tunnel-info API bind host configurable (--tunnel-info-host / GO_IOS_AGENT_HOST)

aluedeke:
## Problem

`ServeTunnelInfo` hard-codes its HTTP listener to `127.0.0.1`:

```go
http.ListenAndServe(fmt.Sprintf("127.0.0.1:%d", port), mux)
```

So the tunnel-info API (`/tunnel/{udid}`, `/tunnels`, `/health`) can only be reached from the same host/network namespace as the daemon. This is a problem when `ios tunnel start` runs in its own container/netns and another process needs to query or stop tunnels.

A host knob already exists for the **client** side — `--tunnel-info-host` is read in `main.go` and `GO_IOS_AGENT_HOST` is honoured by `ios.HttpApiHost()` — but it never reached the **server**, and `--tunnel-info-host` was not even declared in the usage block, so `arguments.String` always errored and silently fell back to the env/default.

## Change

- Thread the host through `startTunnel` → `ServeTunnelInfo`, and bind `fmt.Sprintf("%s:%d", host, port)`.
- Declare `--tunnel-info-host=<host>` in the docopt usage and in `internal/clihelp/help.yaml` (golden files regenerated with `go test . -run TestHelp -update`).
- Fall back to `ios.HttpApiHost()` (`GO_IOS_AGENT_HOST`, default `127.0.0.1`) when the flag is unset/empty — now that the flag is declared, `arguments.String` returns `""` rather than an error, so the empty case is handled explicitly.

**Default behaviour is unchanged** — it still binds `127.0.0.1` unless `--tunnel-info-host` / `GO_IOS_AGENT_HOST` is set. Set e.g. `--tunnel-info-host=0.0.0.0` to expose it.

## Testing

- `go test .` and `go test ./ios/tunnel/` pass; `go vet` and `gofmt` clean.
- `go build` + `ios --help` shows the new flag.

## Note

Exposing the tunnel-info API off loopback also exposes `DELETE /tunnel/{udid}` (stop a tunnel), so it should only be bound on a trusted network — reflected in the help text.


danielpaulus (approved):
Good stuff! Thanks!

PR #754, opened 2026-06-08
fix(forward): stop logging normal connection teardown as errors

aluedeke:
## Problem

The two `io.Copy` goroutines in `StartNewProxyConnection` (`ios/forward/forward.go`) log `forward: close clientConn ...` at **error** level unconditionally — without even inspecting the `io.Copy` return value. Since `io.Copy` returns `nil` on a clean EOF, **every** forwarded connection that opens and closes emits two spurious error lines (one per direction):

```
error  forward: close clientConn --> deviceConn
error  forward: close clientConn <-- deviceConn
```

For workloads that cycle many short-lived forwarded connections this floods logs with errors that arent errors, and makes real failures hard to spot.

## Fix

The two copy goroutines were near-identical, differing only in direction and log message. Factor the half-duplex copy + teardown into a single `copyAndClose` closure that captures the shared connection state, and make its logging level-aware:

- nil / `io.EOF` / `net.ErrClosed` (the "use of closed network connection" you get when the peer goroutine closes the socket) → logged at **debug** as normal lifecycle.
- a genuine copy failure → still logged at **error**, now including the actual `error` value (which the previous code didnt capture at all).

No behavioral change to the forwarding itself — only deduplication, log severity, and the added error detail.

## Testing

- `gofmt` clean
- `go vet ./ios/forward/` clean
- `CGO_ENABLED=0 go build ./ios/forward/` OK


danielpaulus (approved):
Good stuff! Thanks!

Issue #768, opened 2026-06-10
Make ios list discover tunnel/RSD-backed devices


danielpaulus:
**Is your feature request related to a problem? Please describe.**

`ios list` currently appears to be tied to local usbmuxd-style device discovery. That means it returns no devices on a Mac that has a reachable iOS 17+ RSD/userspace tunnel but no locally attached usbmuxd device.

Concrete repro from a remote-device bridge setup:

- office01 has the real paired iPhone locally and runs go-ios userspace tunnel.
- The tunnel agent exposes the device at:
  - UDID: `00008120-001905DE2E9B401E`
  - address: `fd6f:7edb:be57::1`
  - RSD port: `62173`
  - userspace tunnel forwarded locally as `127.0.0.1:61246`
- From another Mac, the tunnel itself is reachable and Xcode/CoreDevice can discover a bridged RemotePairing device.
- But go-ios CLI discovery still reports no local devices:

```sh
go run . list --udid=00008120-001905DE2E9B401E --address=fd6f:7edb:be57::1 --rsd-port=62173 --userspace-port=61246 --details --pretty
```

Observed output:

```json
{
  "deviceList": []
}
```

And commands such as `rsd ls`/`info` fail early with variants of:

```text
Device not found: 00008120-001905DE2E9B401E
Device '00008120-001905DE2E9B401E' not found. Is it attached to the machine?
```

This makes the CLI unusable from a host that can reach an iOS 17+ RSD tunnel but does not have the device in local usbmuxd discovery.

**Describe the solution you'd like**

Make `ios list` and device resolution understand non-usbmuxd tunnel-backed devices.

Desired behavior:

- `ios list` should include devices known from a configured/running go-ios tunnel agent (`/tunnels`) even when usbmuxd has no local attachment.
- Commands should support an explicit hard target using `--address`, `--rsd-port`, and optionally `--userspace-port`, without first requiring the device to resolve through usbmuxd.
- When `--udid` is provided together with explicit tunnel coordinates, use the UDID as identity metadata/filtering, not as a reason to fail if usbmuxd does not know the device.
- The output should make the transport clear, e.g. `connectionType: userspaceTunnel` or similar.

Implementation shape that would fit the current architecture:

- Introduce a device resolver source for RSD/tunnel devices alongside usbmuxd.
- Query the go-ios tunnel agent (`/tunnels`) when available and merge those entries into list output.
- Add a direct RSD target path for commands where `--address --rsd-port` are supplied.
[...]

PR #770, opened 2026-06-11
feat(list): merge usbmux + tunnel-agent device discovery (Phase 1)

danielpaulus:
Phase 1 of making `ios list` show devices beyond local usbmuxd (toward the Xcode-style discovery in #768).

## What
`ios list` now **merges device sources, deduped by UDID**, into a per-device `transports[]` view — so a device reachable both over USB and over an iOS 17+ RSD tunnel shows as **one entry with both transports**.

```jsonc
// ios list   (default JSON)
{"devices":[
  {"udid":"00008110-…","transports":[
    {"type":"usb","source":"usbmux"},
    {"type":"tunnel","source":"tunnel-agent","address":"fdab:c699:7737::1","rsdPort":50507}
  ]}
]}
```

- **`--nojson`** → human table: `UDID  MODEL  OS  VIA  ADDRESS`
- **`--usbmuxd-only`** → legacy behaviour + original `{"deviceList":[…]}` output (compat escape hatch)
- **`--details`** → best-effort lockdown enrichment (model/version) for usbmux-reachable devices

Sources merged this phase: **usbmuxd** (USB + Network entries) and the **go-ios tunnel agent `/tunnels`** registry. The agent fetch is best-effort — a missing agent is silently skipped and never fails the command.

## Validated on real hardware
On a host with one iPhone on USB *and* an active tunnel, plus two pre-iOS17 phones:
```
UDID                       MODEL       OS      VIA         ADDRESS
00008110-001C19A4342040…   iPhone14,2  18.4.1  usb,tunnel  fdab:c699:7737::1:50507
5b98c87e74257…             iPhone8,4   14.3    usb         -
8a4b8ae031a05…             iPhone6,2   12.5.7  usb         -
```
Latency is unchanged vs. legacy (~0.5s; the localhost agent GET adds ~ms).

## Not in this PR (Phase 2)
Native **Bonjour/mDNS** discovery of `_remoted._tcp` / `_remotepairing._tcp` / `_apple-mobdev2._tcp` — the Xcode-style wifi/RSD devices not known to any go-ios agent. That's the bigger lift (adds an mDNS dependency + a bounded browse window, likely a warm-registry daemon for speed) and lands next.

## Tests
Pure `mergeDiscoveredDevices` (usb/network/tunnel/both-sources/dedup/sort) and `renderDeviceTable` are unit-tested. `go build`/`go vet`/`go test -race ./...` green.

### DIFF
--- a/cli_setup.go
+++ b/cli_setup.go
@@ -62,7 +62,7 @@ func configureCLI(arguments docopt.Opts) {
 	slog.Debug("parsed arguments", "args", redactArgs(arguments))
 
 	startAgentFromEnvironment()
-	warnIfAgentIsNotRunning()
+	warnIfAgentIsNotRunning(arguments)
 }
 
 func redactArgs(arguments docopt.Opts) map[string]interface{} {
@@ -87,7 +87,10 @@ func startAgentFromEnvironment() {
 	}
 }
 
-func warnIfAgentIsNotRunning() {
+func warnIfAgentIsNotRunning(arguments docopt.Opts) {
+	if isDeviceListCommand(arguments) {
+		return
+	}
 	if !tunnel.IsAgentRunning() {
 		slog.Warn("go-ios agent is not running. You might need to start it with 'ios tunnel start' for ios17+. Use ENABLE_GO_IOS_AGENT=user for userspace tunnel or ENABLE_GO_IOS_AGENT=kernel for kernel tunnel for the experimental daemon mode.")
 	}
--- a/cmd_global.go
+++ b/cmd_global.go
@@ -67,5 +67,6 @@ func runDeviceListCommand(ctx commandContext) {
 		printDeviceList(details) // legacy path, unchanged output
 		return
 	}
-	printMergedDeviceList(details, tunnelInfoConfigFromArgs(ctx.Args))
+	adhoc, _ := ctx.Args.Bool("--adhoc")
+	printMergedDeviceList(details, adhoc, tunnelInfoConfigFromArgs(ctx.Args))
 }
--- a/cmd_global.go
+++ b/cmd_global.go
@@ -67,7 +67,5 @@ func runDeviceListCommand(ctx commandContext) {
 		printDeviceList(details) // legacy path, unchanged output
 		return
 	}
-	adhoc, _ := ctx.Args.Bool("--adhoc")
-	wifi, _ := ctx.Args.Bool("--wifi")
-	printMergedDeviceList(details, adhoc, wifi, tunnelInfoConfigFromArgs(ctx.Args))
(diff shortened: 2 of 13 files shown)


stoktamisoglu:
Hey @danielpaulus , 

Perhaps you're already aware, just a quick review of the unit tests failures which may help.

The e2e failures on this PR (both Linux and macOS) looks tracing back to one regression.

ios list's default JSON output changed shape, breaking existing e2e assertions

runDeviceListCommand now routes the default ios list (i.e. without --usbmuxd-only) through printMergedDeviceList, which emits:

`{"devices":[{"udid":"...","productType":"...","transports":[{"type":"usb","source":"usbmux"}]}]}`
but `test/e2e/list_test.go` and `test/e2e/preios17/info_test.go` still unmarshal the old shape:


```
var v struct { DeviceList []string `json:"deviceList"` }         // TestList
var v struct { DeviceList []map[string]any `json:"deviceList"` } // TestListDetails — keys "Udid", "ConnectionType", ...
```
Since the top-level key changed (deviceList → devices) and per-device fields changed shape/casing (no ConnectionType; it's now nested under transports[].type), `TestList/TestListDetails` fail for every device on every run:

`list_test.go:22: device 00008110-001C19A43420401E not present in list: []
`confirmed identically in the Linux and macOS` test/e2e `runs and the Linux `test/e2e/preios17` run.

Given --usbmuxd-only already exists as a stated "compat escape hatch," it might be worth flipping the default the other way for this phase (merged discovery opt-in via a flag) until the e2e suite/fixtures are updated for the new schema — otherwise these two test files need updating to the new devices/transports[] shape.

Thx.

danielpaulus:
Thanks for this — merging usbmux + tunnel discovery in `ios list` is genuinely useful. From a CI triage pass, the red real-device e2e here is a real (fixable) contract mismatch, not infra:

This PR changes `ios list --details` to the new nested shape `{"devices":[{"udid":...,"transports":[...]}]}`, but the e2e helpers in `test/e2e/list_test.go` and `test/e2e/info_test.go` still parse the **old flat schema**. As a result `deviceList()` returns `[]`, so `list_test.go:22: device <udid> not present in list: []` fails first and then cascades into essentially every downstream test that resolves a device (WDA, DeviceKit, Info, Ps, RsdLs, FileLs…), on **both** macOS and Linux.

Two things to land this:
1. Update the e2e list/info helpers (and any other `deviceList()` consumer) to the new nested schema.
2. The branch is ~2 months behind `main` now — a rebase will also pick up the newer e2e suite.

If the new schema is meant to be opt-in (to avoid breaking existing JSON consumers), that's worth calling out explicitly too. Glad to re-run the farm once it's updated.

PR #774, opened 2026-06-17
Add --timezone to ios prepare, fix SetSystemTime timezone and ios date output

aluedeke:
## Summary

- Adds `--timezone=<tz>` flag to `ios prepare` so a caller can pin an IANA timezone during device provisioning. Defaults to the host timezone when omitted. Setting timezone via lockdown is only honoured by iOS during the setup-assistant phase; the flag is a no-op on an already-prepared device.
- Improves `ios date`: plain text now prints RFC3339 in the device's own timezone (was RFC850 in host local time); JSON gains a `TimeZone` key and keeps `formatedDate`.
- Adds `ios.SystemTimezone()` and uses it in `SetSystemTime` instead of the hardcoded `Europe/Berlin`.

## API compatibility

`Prepare` gains a `timezone string` parameter (empty means host timezone, like `locale`/`lang`) — this is a source-breaking change for library callers. `SetTime` and `SetSystemTime` signatures are unmodified. `ios date` JSON keeps `formatedDate`, so CLI consumers are unaffected.

## Resolving the host timezone

`time.Local.String()` returns Go's zone *name* (`"Local"` on macOS, `"UTC"`, or an abbreviation), never an IANA name, so it cannot be written to the device. `ios.SystemTimezone()` resolves the real name from, in order: `$TZ`, the `/etc/localtime` symlink target (handles the macOS `/var/db/timezone/tz/<version>/zoneinfo/...` chain), `/etc/timezone` for Debian-style images that ship a plain `/etc/localtime` copy, and finally `UTC`. Covered by unit tests in `ios/system_timezone_test.go`.

## Notes

Attempts to set timezone on an already-prepared iOS 18 device were explored and both available paths fail:
- Lockdown `SetValue("TimeZone")` is silently ignored.
- MDM Settings command over MCInstall returns `DMCSettingsErrorDomain` 28005 ("The time zone could not be set.") — device syslog shows `mc_mobile_tunnel` hitting a sandbox restriction when it tries to call `com.apple.commcenter.coretelephony.xpc` as part of the timezone code path.

Setting timezone during `prepare` (confirmed working on iOS 18.7.1) is the only programmatic path available without a full OTA MDM enrollment.

## Test plan

- [x] `go test ./...` passes (`TestUsesProxy` in `ios/imagemounter` fails on my machine before and after this change — it needs a device without a DDI already mounted)
- [x] `ios date` prints RFC3339 in device timezone; JSON has `TimeIntervalSince1970`, `TimeZone` and `formatedDate`
- [x] `ios prepare --timezone=America/Chicago` pins timezone during provisioning
- [x] `ios prepare` (no `--timezone`) defaults to the host timezone
- [ ] `/test-devices` for the updated e2e `TestDate` assertions


danielpaulus (changes requested):
Thanks @aluedeke! Fixing the hardcoded `Europe/Berlin` is a real improvement and the `ios date` timezone enrichment is nice. I reviewed + tested on a real device; a couple of blockers before this can merge, then it's close.

### 1. 🐞 `time.Local.String()` returns `"Local"`, not an IANA name (confirmed on macOS)
Both `SetSystemTime` (`ios/lockdown_time_zone.go`) and `Prepare`'s default (`ios/mcinstall/prepare.go`) use `time.Local.String()` as the IANA timezone written to the device. I ran it here:
```
macOS, normal env:        time.Local.String() = "Local"
container-like empty env: time.Local.String() = "Local"
TZ= (explicitly empty):   time.Local.String() = "UTC"
```
So on the most common platform for iOS tooling this writes the **invalid** timezone `"Local"` to the device via `SetValueForDomain("TimeZone", "Local")` — strictly worse than the old hardcoded `Europe/Berlin`, which was at least a valid IANA name. `time.Local.String()` returns Go's zone *name* ("Local"/"UTC"/an abbreviation), not a zoneinfo path. To get the real system IANA zone you generally have to resolve the `/etc/localtime` symlink target (or read `$TZ` when set). Worth a small helper with a safe fallback.

### 2. 🔴 This breaks the existing `TestDate` e2e tests
`test/e2e/info_test.go:65` and `test/e2e/preios17/info_test.go:64` both do `smokeObj(t, udid, []string{"TimeIntervalSince1970", "formatedDate"}, "date")` and then parse `formatedDate` as RFC850. This PR removes the `formatedDate` JSON key and switches plain text to RFC3339, but doesn't update these tests — so the real-device suite fails at the required-key check. Please update both tests to assert the new `TimeZone` key / RFC3339 output.

### 3. Breaking change to default `ios date` JSON output
JSON is the default output, and the PR drops `formatedDate` (replacing it with `TimeZone`). Any consumer parsing `formatedDate` breaks. Consider **keeping `formatedDate`** alongside the new `TimeZone` key for back-compat (and updating the tests to match), or at minimum calling this out in release notes.

### 4. `ios date` can now hard-fail on a read
The plain-text path does `time.LoadLocation(deviceTimeZone)` + `exitIfError`, so a device returning an empty or non-IANA `TimeZone` (or the `"Local"` value from #1) makes this read-only command exit 1 where it previously always printed a date. On my test device it returned a valid `Europe/Berlin` so the happy path works — but consider falling back to UTC/local formatting instead of exiting when `LoadLocation` fails.

### Minor
- **Title/scope:** the PR is titled "Add date set command" but there's no `date set` subcommand in the diff — either the title should change or the command is missing.
- `Prepare(..., timezone ...string)` is variadic while the sibling `locale`/`lang` are plain strings with empty-means-default; a plain `timezone string` would be more consistent (and the variadic silently drops a 2nd arg). There's also a pre-existing dead `if err != nil {}` right above the changed `SetTime` call you could drop while you're there.

Once #1 and #2 are sorted I'm happy to run `/test-devices` and get this in. Thanks again! 🙏

aluedeke:
Thanks for the thorough review @danielpaulus — all four points addressed, plus the minors.

**1. `time.Local.String()` returns "Local"** — you're right, and it hit both call sites. Added `ios.SystemTimezone()` (`ios/system_timezone.go`), which resolves the real IANA name from, in order: `$TZ` → the `/etc/localtime` symlink target → `/etc/timezone` → `UTC` as the safe fallback. The symlink resolution handles the macOS chain (`/etc/localtime` → `/var/db/timezone/zoneinfo/Europe/Berlin` → `/private/var/db/timezone/tz/<version>/zoneinfo/Europe/Berlin`) and `zoneinfo.default` variants; every candidate is validated with `time.LoadLocation` before being returned. Reproducing your table on my Mac:

```
normal env:   time.Local.String() = "Local"   ios.SystemTimezone() = "Europe/Berlin"
empty env:    time.Local.String() = "Local"   ios.SystemTimezone() = "Europe/Berlin"
```

Unit tests in `ios/system_timezone_test.go` cover the env/symlink/`/etc/timezone`/fallback paths, the macOS versioned-zoneinfo path shape, and a guard that `SystemTimezone()` always returns something `LoadLocation` accepts.

**2. Broken `TestDate` e2e tests** — updated both `test/e2e/info_test.go` and `test/e2e/preios17/info_test.go`: they now require `TimeZone` in addition to `TimeIntervalSince1970`/`formatedDate`, and assert `TimeZone` is a loadable IANA name on top of the existing RFC850 parse. Compile-checked with `go test -tags e2e -run XXX ./test/...`.

**3. Breaking JSON change** — `formatedDate` is back alongside `TimeZone`, so existing consumers keep working. Both it and the plain-text output are now rendered in the device's timezone rather than the host's, which is the actual behaviour change; the format itself is unchanged (RFC850).

**4. `ios date` hard-failing on a read** — no longer exits. If `LoadLocation` fails it logs a warning and formats in the host timezone, so the command always prints a date.

**Minors:**
- Retitled the PR — there is no `date set` subcommand, the title was stale.
- `Prepare` now takes a plain `timezone string` (empty means host timezone), consistent with `locale`/`lang`. Note this is source-breaking for library callers, unlike the variadic — flagging in case you'd rather I keep it variadic.
- Dropped the dead `if err != nil {}` above the `SetTime` call.

Ready for `/test-devices` whenever you are 🙏

danielpaulus:
Re-reviewed after the fix commit (`a4703380`). Thanks for the thorough turnaround, @aluedeke — this is a solid pass and the prior blockers are genuinely resolved:

- **#1 (`time.Local.String()` → "Local")** — fixed. `ios.SystemTimezone()` (`ios/system_timezone.go`) resolves the real IANA name via `$TZ` → `/etc/localtime` symlink → `/etc/timezone` → `UTC`, validating every candidate with `time.LoadLocation`. The `zoneNameFromPath` handling of the macOS versioned/`zoneinfo.default` paths is a nice touch, and the unit tests in `ios/system_timezone_test.go` cover the env/symlink/file/fallback matrix well — including a guard that the result is always `LoadLocation`-able.
- **#2 (broken `TestDate` e2e)** — fixed. Both `test/e2e/info_test.go` and `test/e2e/preios17/info_test.go` now require `TimeZone` and assert it's a loadable IANA name on top of the RFC850 parse.
- **#3 (dropped `formatedDate`)** — fixed. `formatedDate` is back alongside `TimeZone` in the JSON, so existing consumers keep working; it's now rendered in the device zone (the intended behavior change).
- **#4 (read hard-failing)** — fixed. `printDeviceDate` (`main.go:1277`) now warns and falls back to the host zone instead of exiting.
- Minors (title, plain `timezone string`, dead `if err != nil {}`) all addressed.

Local status: `go build ./...`, `go vet ./...`, and `go test ./...` all pass. CI: lint / linux / windows green; real-device is skipping pending `/test-devices`.

Two remaining items, neither a hard blocker but worth a look before device-testing:

1. **No validation of user-supplied `--timezone` before it's written to the device** (`ios/lockdown_time_zone.go:20`, reached via `ios/mcinstall/prepare.go:113`). The *read* path (`main.go:1277`) was hardened with `LoadLocation`, but the *write* path — the new user-facing feature — isn't: an invalid `--timezone=garbage` flows straight into `SetTime`, which first sets the clock (`TimeIntervalSince1970`) and only then attempts the bad `TimeZone` write, so a typo can leave the clock changed and either error out mid-way or get silently accepted by lockdown. A one-line `time.LoadLocation(timezone)` guard in `runPrepareCommand` (or at the top of `Prepare`, after the empty→host default) that fails fast before any device mutation would make this robust and give the user a clear error. `SystemTimezone()` is already always valid, so this only affects the explicit-flag case.

2. **`ios date` plain-text (non-JSON) format changed RFC850 → RFC3339** (`main.go:1284`). The JSON path preserves `formatedDate` (RFC850), but the `--nojson` path now prints RFC3339, which is a visible change for anyone scripting against `ios date --nojson`. Fine if intentional (RFC3339 is arguably the better machine format), but worth either keeping RFC850 for back-compat or calling it out in the release notes.

Minor/optional: on Windows with `TZ` unset, `SystemTimezone()` returns `UTC` since detection is `/etc`-based (`ios/system_timezone.go`), so `prepare`'s "defaults to the host timezone" isn't quite true there — not a correctness bug (UTC is valid), just a doc nuance. `SetSystemTime` is now unused in-tree but stays as a valid exported library helper.

Recommendation: in good shape. I'd suggest adding the `--timezone` validation (item 1) since it directly guards the new flag against partial device state; item 2 is your call (keep-or-document). Once you're happy, this looks ready for a maintainer to run `/test-devices`. Not approving/merging — leaving that to @danielpaulus.

_Automated review pass._

danielpaulus:
Merging — thank you, @aluedeke! 🙏 The fix commit resolved all the earlier feedback: the real IANA `SystemTimezone()` resolver (with $TZ → /etc/localtime → /etc/timezone → UTC, each validated) replaces the `time.Local.String()` → "Local" bug, the e2e assertions and `formatedDate` are restored, and `ios date` now warns + falls back instead of hard-failing. Verified it builds and the full suite passes after merging current main (the `--timezone` usage line merged cleanly into the relocated `cliUsage()` from #796).

Two small follow-ups on record (non-blocking, happy to take a quick PR or do them myself if you'd like):
1. **Validate `--timezone` before mutating the device** — the write path (`ios/lockdown_time_zone.go` via `prepare.go`) sets the clock first then writes the TZ, so an invalid IANA name can leave partial state; a one-line `time.LoadLocation` guard up front would fail fast.
2. **`ios date --nojson` plain-text format changed RFC850 → RFC3339** — JSON `formatedDate` stays RFC850, but anything parsing the plain-text output would see the change; flagging for awareness.

Thanks again for the fix and the follow-through! 🚀

danielpaulus (approved):
Prior feedback all addressed; verified build + tests pass on the merged result. Thanks @aluedeke!

PR #776, opened 2026-06-22
fix(imagemounter): apply RestoreRequestRules when personalizing DDI

qiuzilin:
Personalized Developer Disk Image mounting (iOS 17+) requests a signature from Apple's TSS server. Newer BuildManifests (e.g. iOS 26) no longer carry static EPRO/ESEC values on the trusted manifest entries; instead they ship `RestoreRequestRules` (on the LoadableTrustCache entry) that derive these fields from the request parameters.

Previously getSignature read entry.EPRO/entry.ESEC directly, which decode to their zero value (false) for these manifests. Apple then rejects the request with `STATUS=94 (This device isn't eligible for the requested build)` because a production + secured Img4 device must send EPRO=true/ESEC=true.

Parse RestoreRequestRules from the manifest and apply them (matching the algorithm used by idevicerestore / pymobiledevice3) so the correct EPRO/ESEC are sent. Verified end-to-end against a physical iOS 26.5 device, which now mounts successfully.

Adds unit tests for applyRestoreRequestRules.

### DIFF
--- a/ios/imagemounter/personalized_image.go
+++ b/ios/imagemounter/personalized_image.go
@@ -38,6 +38,11 @@ func (m buildManifest) findIdentity(identifiers personalizationIdentifiers) (bui
 	return buildIdentity{}, fmt.Errorf("findIdentity: failed to find identity for ApBoardId 0x%x and ApChipId 0x%x", identifiers.BoardId, identifiers.ChipID)
 }
 
+type restoreRequestRule struct {
+	Actions    map[string]interface{} `plist:"Actions"`
+	Conditions map[string]interface{} `plist:"Conditions"`
+}
+
 type manifestEntry struct {
 	Digest  []byte
 	Trusted bool `plist:"Trusted"`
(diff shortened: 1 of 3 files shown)


shamanec:
Important change, I had to add those properties manually to the manifest so it would work 😅 I'd rather @danielpaulus approve it for merging though

stoktamisoglu:
> Important change, I had to add those properties manually to the manifest so it would work 😅 I'd rather @danielpaulus approve it for merging though

Hi @qiuzilin / @shamanec , do you think we still need this PR after merging #781  ?

shamanec:
I would say we dont as long as it works fine on the latest release 👌

stoktamisoglu:
FYI: have tested #781 on 26.5.2 and it fixed the problem.

qiuzilin:
> > Important change, I had to add those properties manually to the manifest so it would work 😅 I'd rather @danielpaulus approve it for merging though
> 
> Hi @qiuzilin / @shamanec , do you think we still need this PR after merging #781 ?

Thanks @stoktamisoglu for #781, and thanks @shamanec for checking.

Given that #781 is merged and confirmed on 26.5.2, this PR is redundant — I'll close #776 as superseded by #781.

Thanks everyone for the review and for verifying the fix!

PR #778, opened 2026-06-23
feat(mdm): add supervised passcode and Screen Time management commands

aluedeke:
## Summary

Adds three new subcommands under `ios mdm` for supervised device passcode management via the MCInstall escalation channel:

- **`ios mdm fetch-unlock-token`** — saves the device passcode unlock token to a file; must be run while the device has no passcode set (typically during provisioning)
- **`ios mdm clear-passcode`** — removes the device lock passcode using a previously saved unlock token; works regardless of current lock state and without knowing the passcode
- **`ios mdm clear-screen-time-password`** — clears the Screen Time restrictions PIN; no token required, supervisor identity alone suffices

All three commands require a supervised device and a PKCS#12 supervisor identity (`--p12file` + `--password` or `P12_PASSWORD` env var).

## Protocol background

The device lock passcode flow is a two-step MCInstall protocol (discovered by reverse engineering `MobileDeviceKit.framework` from Apple Configurator 2):

1. `RequestUnlockToken` → device returns a ~1300-byte keybag escrow blob (only works when no passcode is set)
2. `ClearPasscode` + `UnlockToken` → clears the passcode (works even when device is locked)

The Screen Time command is a single MCInstall `ClearRestrictionsPassword` request with no additional token required.

## New MCInstall methods

- `FetchUnlockToken() ([]byte, error)` — on an already-escalated connection
- `FetchUnlockTokenSupervised(p12bytes, password)` — escalates then fetches
- `ClearPasscodeSupervised(p12bytes, password, token)` — escalates then clears
- `ClearScreenTimePasswordSupervised(p12bytes, password)` — escalates then clears Screen Time PIN

## Test plan

- [ ] `ios mdm fetch-unlock-token --p12file=<file> --password=<pw> --output=/tmp/token.bin` on a passcode-free supervised device → file written, JSON `{"bytes":1304,"path":"..."}` printed
- [ ] Set a passcode on the device, then `ios mdm clear-passcode --p12file=<file> --password=<pw> --token=/tmp/token.bin` → `{"status":"ok"}`, passcode gone
- [ ] `ios mdm fetch-unlock-token` on a device with passcode set → exits 1 with `DMCKeybagErrorDomain` error
- [ ] `ios mdm clear-screen-time-password` on a device with Screen Time PIN → `{"status":"ok"}`, PIN cleared

### DIFF
--- a/cmd_device_management.go
+++ b/cmd_device_management.go
@@ -336,6 +336,13 @@ func runMdmCommand(ctx commandContext) {
 		return
 	}
 
+	if securityInfo, _ := ctx.Args.Bool("security-info"); securityInfo {
+		info, err := conn.SecurityInfo()
+		exitIfError("failed to fetch security info", err)
+		fmt.Println(convertToJSONString(info))
+		return
+	}
+
 	if clearScreenTime, _ := ctx.Args.Bool("clear-screen-time-password"); clearScreenTime {
 		err = conn.ClearScreenTimePassword()
 		exitIfError("failed to clear Screen Time password", err)
--- a/cmd_device_management.go
+++ b/cmd_device_management.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"strings"
@@ -129,10 +130,7 @@ func runPrepareCommand(ctx commandContext) {
 func runSetWallpaperCommand(ctx commandContext) {
 	imagePath, _ := ctx.Args.String("<imagePath>")
 	p12file, _ := ctx.Args.String("--p12file")
-	p12password, _ := ctx.Args.String("--password")
-	if p12password == "" {
-		p12password = os.Getenv("P12_PASSWORD")
-	}
+	p12password := p12PasswordOrEnv(ctx)
 	screen, _ := ctx.Args.String("--screen")
 	if screen == "" {
 		screen = "home"
@@ -173,10 +171,7 @@ func runHTTPProxyCommand(ctx commandContext) {
 		pass = os.Getenv("PROXY_PASSWORD")
 	}
 	p12file, _ := ctx.Args.String("--p12file")
-	p12password, _ := ctx.Args.String("--password")
-	if p12password == "" {
-		p12password = os.Getenv("P12_PASSWORD")
-	}
+	p12password := p12PasswordOrEnv(ctx)
 	p12bytes, err := os.ReadFile(p12file)
 	exitIfError("could not read p12-file", err)
 
@@ -192,10 +187,7 @@ func runProfileCommand(ctx commandContext) {
 	if add, _ := ctx.Args.Bool("add"); add {
 		name, _ := ctx.Args.String("<profileFile>")
 		p12file, _ := ctx.Args.String("--p12file")
-		p12password, _ := ctx.Args.String("--password")
-		if p12password == "" {
-			p12password = os.Getenv("P12_PASSWORD")
-		}
+		p12password := p12PasswordOrEnv(ctx)
 		if p12file != "" {
 			handleProfileAddSupervised(ctx.Device, name, p12file, p12password)
 			return
@@ -214,10 +206,7 @@ func runDeviceNameCommand(ctx commandContext) {
 
 func runPairCommand(ctx commandContext) {
 	org, _ := ctx.Args.String("--p12file")
-	pwd, _ := ctx.Args.String("--password")
-	if pwd == "" {
-		pwd = os.Getenv("P12_PASSWORD")
-	}
+	pwd := p12PasswordOrEnv(ctx)
 	pairDevice(ctx.Device, org, pwd)
 }
 
@@ -272,18 +261,42 @@ func runDevModeCommand(ctx commandContext) {
 	}
 }
 
+// p12PasswordOrEnv returns the supervision identity password from --password, falling
+// back to the P12_PASSWORD environment variable.
+func p12PasswordOrEnv(ctx commandContext) string {
+	p12password := p12PasswordOrEnv(ctx)
+	return p12password
+}
+
+// decodeBase64Flexible decodes a base64 token that may arrive from a secrets manager
+// with surrounding whitespace, hard line wrapping, no padding, or in the URL-safe
+// alphabet. Any of those silently truncated or rejected the token previously.
+func decodeBase64Flexible(encoded string) ([]byte, error) {
+	compact := strings.Join(strings.Fields(encoded), "")
+	if compact == "" {
+		return nil, fmt.Errorf("no token data received")
+	}
+	for _, encoding := range []*base64.Encoding{
+		base64.StdEncoding, base64.RawStdEncoding,
+		base64.URLEncoding, base64.RawURLEncoding,
+	} {
+		if decoded, err := encoding.DecodeString(compact); err == nil {
+			return decoded, nil
+		}
+	}
+	return nil, fmt.Errorf("not valid base64 (tried standard and URL-safe, padded and raw)")
+}
+
 func runMdmCommand(ctx commandContext) {
 	p12file, _ := ctx.Args.String("--p12file")
-	p12password, _ := ctx.Args.String("--password")
-	if p12password == "" {
-		p12password = os.Getenv("P12_PASSWORD")
-	}
+	p12password := p12PasswordOrEnv(ctx)
 
 	p12bytes, err := os.ReadFile(p12file)
 	exitIfError("could not read p12file", err)
 
 	conn, err := mcinstall.New(ctx.Device)
 	exitIfError("failed to connect to MCInstall service", err)
+	defer conn.Close()
 
 	err = conn.Escalate(p12bytes, p12password)
 	exitIfError("failed to escalate MCInstall session", err)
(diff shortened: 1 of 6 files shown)


danielpaulus:
Heads-up on testing: these commands require a **supervised device with a known supervisor identity**, which the CI farm doesn't have yet — I filed #789 to add one. Until then the `mdm` commands can't get real-device e2e coverage, and the `PairSupervised` locked-device path is only partially exercised by the existing pairing tests.

I also ran a multi-agent review (posting the findings as a separate review). The short version: the feature is well-built and idiomatic, but there are a couple of things worth addressing before merge — most importantly the unlock token can end up in logs in plaintext on an error path, and the locked-device pairing change currently reports success without persisting a pair record. Details in the review.

danielpaulus (changes requested):
Thanks @aluedeke — this is a genuinely useful feature and the code is clean and idiomatic (logging, docopt wiring, and error style all match the codebase; the panic-fix in `PairSupervised` is a real improvement). I ran a multi-agent review; a few things worth addressing before merge, roughly in priority order. As noted in #789 we don't have a supervised device in CI yet, so I couldn't exercise these on hardware — you likely can on yours.

---

### 1. 🔒 The unlock token can leak into logs in plaintext
`ios/mcinstall/mcinstall.go` — `FetchUnlockToken` (`RequestUnlockToken failed: %+v`) and `ClearPasscode` (`ClearPasscode failed: %+v`) format the **entire device response** into the error on any non-`Acknowledged` status. For `RequestUnlockToken` that response *is* the token carrier, and for `ClearPasscode` the device commonly echoes the `UnlockToken` you sent. The error then flows `exitIfError → logFatal → slog.Error(msg, "err", err)`, so the raw token bytes land in stderr/log sinks — which defeats the careful `0o600` file permission the feature otherwise uses to protect it.

Fix: don't `%+v` the whole response on these paths. Surface only `Status` / `ErrorChain` (redacting `UnlockToken`). This also lets you deliver on the doc-comment promise (see #6).

### 2. Locked-device pairing reports success without persisting a pair record
`ios/pair.go` — when the device is locked it returns neither `EscrowBag` nor a string `Error`, so `if !ok { return nil }` returns success. But `savePair` is skipped, so no usbmux pair record is written. Both callers then claim full success: `main.go` logs `Successfully paired <serial>` and the REST endpoint (`restapi/api/device_endpoints.go`) returns HTTP 200 `Device paired`. A user/automation then proceeds to use services needing a valid pair record and hits confusing downstream failures.

The panic-fix itself is right — the issue is calling it *success*. Suggest returning a distinct sentinel/typed error (e.g. `ErrDeviceLockedPairingDeferred`) or at minimum a clearly different log so callers don't report a pairing that didn't happen.

### 3. The `Error` guard only catches string values
`ios/pair.go` — `respMap["Error"].(string)` only matches when the error is a string. This protocol encodes some failures as maps (e.g. `ExtendedResponse` elsewhere in the pairing code); such a rejection plus a missing `EscrowBag` falls straight through to the silent-`nil` success path, swallowing a genuine failure. A broader "no EscrowBag ⇒ not a clean success" check covers both this and #2.

### 4. Missing `defer conn.Close()`
`cmd_device_management.go` — `runMdmCommand` never closes the `mcinstall` connection after `mcinstall.New` succeeds; every other `mcinstall.New` call site pairs with a `Close`. Harmless for the one-shot CLI, but this handler is the natural template for the REST API, where a long-lived process would leak the escalated MCInstall session/socket per call.

### 5. `--token=-` stdin parsing is fragile
`cmd_device_management.go` — `fmt.Scan(&encoded)` reads only the first whitespace-delimited field, so a token piped from a secrets manager with a trailing newline + metadata (or line-wrapping) is silently truncated. Also `base64.StdEncoding` rejects URL-safe (`-`/`_`) tokens. Prefer `io.ReadAll(os.Stdin)` + `strings.TrimSpace`, and consider accepting base64url.

### 6. Doc-comment promises an error signal the code never extracts
`ios/mcinstall/mcinstall.go` — `FetchUnlockToken`'s comment says it returns a `DMCKeybagErrorDomain` code `37002` error when a passcode is active, but `checkStatus` only tests `Status == "Acknowledged"`, so that case just becomes the opaque `%+v` dump. Fixing #1 (surface `ErrorChain`) is the natural place to make the comment true and give users the actionable "remove the passcode first" reason.

---

Minor (optional): the `--password`-or-`P12_PASSWORD` block is now duplicated ~6× in `cmd_device_management.go` — a small `p12PasswordOrEnv(ctx)` helper would DRY it up; and `FetchUnlockToken`/`ClearPasscode`/`ClearScreenTimePassword` share a `sendAndReceive → checkStatus → Errorf` shape that could collapse into one `sendChecked(req, name)` (the natural home for the #1/#6 redaction).

Happy to run `/test-devices` once #789 gives us a supervised phone. Thanks again — solid contribution, just want the token-handling and pairing-success semantics tightened first. 🙏

aluedeke:
Thanks for the thorough review @danielpaulus — all six items plus the DRY suggestion are addressed and pushed:

- `e79fd80` fix(pair): report locked devices instead of reporting success — items **2, 3**
- `a7f7724` fix(mdm): stop formatting device responses into errors, surface 37002 — items **1, 6**
- `5c65cb1` fix(mdm): close MCInstall session, harden `--token=-` decoding — items **4, 5** + DRY
- `9fa2d3e` fix(mdm): repair `p12PasswordOrEnv` recursing into itself — see the note at the end

Since #789 means CI can't exercise these, I ran them on a supervised device: **iPhone 16 Pro Max (`iPhone17,2`), iOS 26.5, supervised**. Before/after with the same command against the same device state.

### Item 6 (and 1) — 37002 now actionable, response no longer dumped

Passcoded device, valid pair record, `ios mdm fetch-unlock-token`:

**Before**
```
err: "RequestUnlockToken failed: map[CommandErrorArchive:[4 11 115 116 114 101 97 109 116 121
112 101 100 129 232 3 132 1 64 132 132 132 7 78 83 69 114 114 111 114 0 132 132 8 78 83 79 98
106 101 99 116 0 133 132 3 105 64 64 130 138 144 0 0 …] ErrorChain:[map[ErrorCode:37002
ErrorDomain:DMCKeybagErrorDomain LocalizedDescription:The operation couldn’t be com…"
```

**After**
```
err: "device has a passcode set; remove it before fetching an unlock token:
RequestUnlockToken failed: Status=Error; DMCKeybagErrorDomain(37002):
The operation couldn’t be completed. (DMCKeybagErrorDomain error 37002.)"
```

Both exit 1. That `CommandErrorArchive` byte array is exactly the vector you flagged — on the paths where the response carries the token, it went to the log verbatim. Callers can now match on `ErrPasscodeSet` instead of grepping a number out of formatted output, which is what makes a "clear the passcode only if one is set" guard practical for us.

### Item 2 — locked-device pairing

Passcoded device, screen locked, no valid pair record (the pair record had just been invalidated by the passcode-set USB reset), `ios pair --p12file`:

**Before**
```
INFO  "Successfully paired 00008140-…"                                   EXIT=0
$ idevicepair validate  ->  "Device 00008140-… is not paired with this host"
```

**After**
```
ERROR "Pairing incomplete for 00008140-…"
      err: "device is locked: supervised pairing was accepted but no pair record
            could be persisted, unlock the device once and pair again"    EXIT=1
```

And it does not over-trigger: with the same passcode still set but the device **unlocked**, `ios pair --p12file` returns `Successfully paired`, exits 0, and `idevicepair validate` succeeds.

This one bit us in production before the fix: a device whose record had been invalidated reported a successful pairing, and the boot then died several steps later at an unexplained `InvalidHostID`. The REST endpoint now answers `423 Locked` for the same condition.

### Item 5 — `--token=-`

`clear-passcode` with the token hard-wrapped to 40-column lines plus trailing newlines (46 lines total) → `{"status":"ok"}`. `fmt.Scan` previously consumed only the first line, and the resulting failure looked like a stale token rather than a truncated one — a genuinely expensive misdiagnosis in our setup.

### Items 3 and 4 — code-only

I could not provoke a map-shaped pairing rejection on demand, so **item 3** is reasoned-about rather than observed; the widened guard makes it a reported rejection instead of falling through to the locked-device branch. **Item 4** (`defer conn.Close()`) has no externally observable behaviour on a one-shot CLI.

### Note on `9fa2d3e`

While extracting `p12PasswordOrEnv` I replaced every occurrence of the `--password`-or-`P12_PASSWORD` block with a call to the new helper — including the occurrence that had just become the helper's own body. It recursed until the stack overflowed, breaking every command that reads a supervision password. `go build` cannot see unbounded recursion and no unit test covers those handlers, so it only surfaced when I ran the binary against the device. Fixed in `9fa2d3e`; I've left it as its own commit rather than rewriting history so the branch stays reviewable, and it squashes away on merge.

Happy to run anything else you'd like checked on real hardware while the device is on the bench — including `/test-devices` scenarios once #789 lands.

danielpaulus (approved):
All six review findings addressed thoroughly — verified commit-by-commit:

1. ✅ Token no longer formatted into errors (`describeFailure` extracts only Status + ErrorChain; all methods route through `sendChecked`)
2. ✅ Locked-device pairing returns `ErrDeviceLockedPairingDeferred` — main.go exits non-zero, REST API returns HTTP 423 instead of 200
3. ✅ `Error` guard now matches any type, not just strings
4. ✅ `defer conn.Close()` added
5. ✅ `decodeBase64Flexible` (strips whitespace, tries std/url × padded/raw) replaces `fmt.Scan`
6. ✅ 37002 surfaced as `ErrPasscodeSet` with an actionable message

Bonus `security-info`/`PasscodePresent` read-only query is a clean addition, and the DRY helpers (`p12PasswordOrEnv`, `sendChecked`) tidy things up nicely. Build/vet/gofmt/unit tests green; real-device e2e passed on both runners (no regressions to normal pairing).

Note: the mdm commands themselves remain device-untested until a supervised device lands (#789). Merging on the strength of the code review + green e2e. Thanks for the thorough turnaround, @aluedeke! 🙏

PR #773, opened 2026-06-17
Implement pasteboard service control

shamanec:
* Set/Get pasteboard values on iOS 17+ devices

Kudos to @doronz88 for reverse-engineering the service

PR #786, opened 2026-08-03
Fix imagemounter to validate mount/unmount reply

stoktamisoglu:
func (p PersonalizedDeveloperDiskImageMounter) mountPersonalizedImage(signatureBytes []byte, trustCache []byte) 

Discards the unsuccessful mounting when device returns nil or status is not 'Complete'. It handles both condition and return error to catch unsuccessful mounting (like developer mode is not enabled etc.)

New tests are also added.

### DIFF
--- a/ios/imagemounter/imagedownloader_test.go
+++ b/ios/imagemounter/imagedownloader_test.go
@@ -59,6 +59,10 @@ func TestUsesProxy(t *testing.T) {
 		t.Skip("No device attached")
 		return
 	}
+	// the device rejects 'MountImage' when an image is already mounted, so start from a clean state
+	if err := imagemounter.UnmountImage(d.DeviceList[0]); err != nil {
+		log.Printf("could not unmount an already mounted developer disk image: %v", err)
+	}
 	wg.Add(1)
 	m, err := imagemounter.NewPersonalizedDeveloperDiskImageMounter(d.DeviceList[0], ios.IOS17())
 	if !assert.Nil(t, err) {
@@ -92,6 +96,10 @@ func TestWorksWithoutProxy(t *testing.T) {
 	if len(d.DeviceList) == 0 {
 		t.Skip("No device attached")
 		return
+	}
+	// the device rejects 'MountImage' when an image is already mounted, so start from a clean state
+	if err := imagemounter.UnmountImage(d.DeviceList[0]); err != nil {
+		log.Printf("could not unmount an already mounted developer disk image: %v", err)
 	}
 	m, err := imagemounter.NewPersonalizedDeveloperDiskImageMounter(d.DeviceList[0], ios.IOS17())
 	if !assert.Nil(t, err) {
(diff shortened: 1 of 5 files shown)


stoktamisoglu:
Hey @shamanec , this one (#786) and #787 are the PRs just fixing better error/return status handling which triggers some confusions in the reported issues. If you could review at your conveniences, I would appreciate. Thx.

danielpaulus (changes requested):
Thanks @stoktamisoglu — validating the mounter replies is absolutely the right direction (we just shipped the same philosophy in #773/#787), and the tests + the openstdio-style structure are appreciated. I ran a multi-agent review and also **tested this branch on a real iOS 14.3 device**, which confirmed a few things that need another pass before merge:

### 1. The pre-iOS17 mount path is still unvalidated (main gap)
`DeveloperDiskImageMounter.mountImage` (imagemounter.go) still writes the `MountImage` command and returns nil without reading the reply — only the personalized (iOS 17+) mounter got the validation. On iOS ≤16 a rejected mount (locked device, bad signature) still prints *success mounting image* and exits 0 — the exact bug class the PR title targets. Please validate the reply there too.

### 2. Surface `Error`/`DetailedError` instead of dumping the map
Verified on a real iPhone (iOS 14.3), `unmount` with nothing mounted replies:
```
{Error: InternalError,
 DetailedError: Error Domain=com.apple.MobileStorage.ErrorDomain Code=-2 "Failed to unmount /Developer."
   … "There is no matching entry in the device map for /Developer." …}
```
The current code prints the entire raw Go map (`unexpected response: map[DetailedError:…]`) with nondeterministic key order. `listImages()` in the same file already has the right pattern — `device responded with error: %v` from `resp["Error"]`. Suggest one shared helper (there are now three copies of the same read-and-check block: the new mount check, `readUnmountResponse`, and the pre-existing `waitForUploadComplete`) that checks `Error` first, then `Status == "Complete"`. Feel free to use the capture above as a realistic test fixture — the invented `ImageNotMounted` / `Status: Failure` shapes in the tests don't match what devices actually send.

### 3. `ios image unmount` still exits 0 on failure
Confirmed on device: with this branch, a failed unmount logs `error unmounting image` but the process exits 0 (`cmd_device_debug.go` unmount branch logs + returns). Now that unmount can actually report failure, it should `os.Exit(1)` like the mount branch one block above (added deliberately in #768) — otherwise the new validation is invisible to scripts/CI.

### Smaller points
- The new `golog.Debug("mount response", …)` is missing the `"udid"` attr (device is in scope — see AGENTS.md; every other log line in that file carries it).
- `encodeUnmountResponse` in unmount_test.go duplicates `encodePlistMessage` in personalized_image_mounter_test.go — one helper can serve both files.
- The additions to `imagedownloader_test.go` extend device-dependent behavior inside package unit tests; per AGENTS.md device-dependent testing belongs in `test/e2e/`. Fine to leave the pre-existing tests, but let's not grow them.
- Good news from the device run: iOS 14.3 does reply `Status: Complete` on successful unmount, so the strict success check itself is fine there. iOS 12 may answer `Error: UnknownCommand` for unmount (it predates the command per pymobiledevice3) — worth tolerating that as a no-op if you want unmount to stay best-effort on very old devices.

Happy to re-run the real-device suite (`/test-devices`) as soon as these are in. Thanks again — this is close!

PR #787, opened 2026-08-03
Improvement/fix(rsd): fail fast when a service is missing from the RSD service list

stoktamisoglu:
GetPort returns 0 for a service that is not in the RSD service list, and the callers dialed that port without checking it. Services provided by the developer disk image are only listed once the image is mounted, so a missing image surfaced as 'connection forcibly closed by the remote host' on the tunnel address, which gives no hint about the actual cause.

Report the missing service by name instead, following the check that getUntrustedTunnelServicePort already does.

New tests are also added.

### DIFF
--- a/ios/connect.go
+++ b/ios/connect.go
@@ -108,13 +108,15 @@ func ConnectToService(device DeviceEntry, serviceName string) (DeviceConnectionI
 	return muxConn.ReleaseDeviceConnection(), nil
 }
 
-// rsdPortForService looks up the port a service is listening on in the RSD service list of the device.
-// Services provided by the developer disk image only show up after it has been mounted. Without this
-// check we would dial port 0 and report a confusing 'connection refused' instead of the actual cause.
-func rsdPortForService(device DeviceEntry, service string) (int, error) {
-	port := device.Rsd.GetPort(service)
+// RsdPortForService looks up the port a service is listening on in the given RSD service list.
+// Services provided by the developer disk image only show up after a recent enough image has been
+// mounted: mounting adds them, but an outdated image can lack newer services even when mounted.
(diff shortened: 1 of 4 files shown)


danielpaulus:
## Thank you, @stoktamisoglu! 🙏

This is a genuinely valuable fix — and the timing couldn't have been better. While reviewing #773 today we hit *exactly* the failure mode this PR eliminates: `com.apple.coredevice.pasteboardservice` missing from the RSD list surfaced as a meaningless `connection refused` on port 0, and it took a live debugging session on the device host to find the real cause (the mounted DDI was too old to provide the service). With your change the error names the service and points at the fix immediately. The tests are exactly right too — device-free, fast, and they pin the message.

The review also confirmed completeness: every `GetPort` dial site in the repo is now guarded, and no caller depended on the old behavior.

I pushed one maintainer commit on top (`83ba032`) extending your work:
- exported the helper as `ios.RsdPortForService(RsdPortProvider, service)` so `openstdio` shares it instead of carrying a copy of the message
- added the "an outdated image can lack this service even when mounted" hint — that was the exact trap we fell into today
- a `port <= 0` backstop in `ConnectTUNDevice` (on the userspace-TUN path a port-0 dial silently "succeeds")
- fixed a previously swallowed RSD handshake error in `main.go` that would have made every service misreport as missing

Kicking off the real-device suite now; merging once green. Thanks again — quality contribution! 🚀

danielpaulus (approved):
Reviewed (8-angle pass, all dial sites covered) and verified on real devices via /test-devices — both runners green. LGTM with the maintainer commit on top.

stoktamisoglu:
### Thank you @danielpaulus. 

Very good additions indeed.

PR #788, opened 2026-08-03
Fix/imagemounter validate replies

stoktamisoglu:
`mountPersonalizedImage` and both `UnmountImage` implementations send a plist command and then discard the device's reply, so a rejected operation is reported as success — on iOS 26.5.2 a mount with Developer Mode off replies `Error:ImageMountFailed ... "Developer mode is not enabled."` and an unmount on a locked device replies `Error:DeviceLocked`, yet go-ios prints `success` and returns `nil` in both cases (the unmounted DDI then surfaces much later as a confusing `connection forcibly closed` from dialing port 0). This validates the replies against `Status == "Complete"`, matching what the sibling `waitForUploadComplete` already does, so the device's own error text reaches the caller. Note the low-level mount call now errors when an image is already mounted — the "already mounted is fine" policy is unchanged in `MountImage(device, path)`, which still pre-checks with `ListImages()` — and the two real-device tests were relying on the swallowed error, so they now unmount first. Covered by unit tests and verified on-device (mount, unmount, and locked-device rejection).


stoktamisoglu:
Hey @shamanec , I updated the mount PR by including both mount and unmount device reply. This PR (#788) and the #787 both improves the device return messages handling to avoid false positive conditions and better logging of the root cause of the issue instead silent discarding. If you could review at your conveniences, I would appreciate. Thx.

stoktamisoglu:
Thanks @danielpaulus for the review of #786 which is followed up here at #788. 

All addressed in the commit bf229ae, and I got hold of an iOS 16.4.1 device so the pre-iOS17 path is now verified rather than assumed.

1. `DeveloperDiskImageMounter.mountImage` now validates its reply too.
2. One shared `readImageMounterResponse` replaces what turned out to be **four** copies — `sendUploadRequest` had one as well. It takes the expected status as a parameter, because `ReceiveBytes` acks with `ReceiveBytesAck`, not `Complete`; consolidating on a single `Complete` would have broken the upload step. Errors now follow the `listImages` shape with `DetailedError` appended.
3. `ios image unmount` exits 1 on failure.

Smaller ones: `udid` on the new log lines (added a `udid` field to `DeveloperDiskImageMounter`, which held no device reference); one shared plist test helper; `imagedownloader_test.go` reverted to its original state per AGENTS.md. `Error: UnknownCommand` on unmount is tolerated as a no-op via a typed `deviceRefusedError`, so pre-iOS13 devices keep the old best-effort behaviour.

Test fixtures are now all real captures: developer-mode-disabled and `DeviceLocked` from iOS 26.5.2, plus your iOS 14.3 unmount reply — the invented shapes are gone.

Verified on device (iPhone 8 / iOS 16.4.1, legacy mounter):

| | exit | output |
|---|---|---|
| v1.2.1 | 0 | `success unmounting image` |
| this branch | 1 | `UnmountImage: device responded with error: InternalError: ... "There is no matching entry in the device map for /Developer."` |

Legacy mount happy path still fine (`image auto` → success, `image list` → signature), and re-checked the personalized path on iOS 26.5.2 (unmount → none → remount → signature) after the refactor.

Thanks again for the review. It is ready for /test-devices I believe.

danielpaulus:
Thanks! This is really great work. @stoktamisoglu

danielpaulus (approved):
All review findings addressed (verified): plain pre-iOS17 mount path now validated, one shared `readImageMounterResponse` replacing four copies (upload two-phase handshake correctly preserved), `ios image unmount` exits 1 on failure, errors follow the `listImages` shape via typed `deviceRefusedError`, real device-captured fixtures, `udid` on log lines, and `UnknownCommand` tolerated as a no-op for pre-iOS13. Nicely done getting an iPhone 8 / iOS 16.4.1 to verify the legacy mounter. Real-device suite green on both runners. Thanks @stoktamisoglu! 🚀

stoktamisoglu:
Thanks for the review and your contribution into this @danielpaulus . When do we expect a new go-ios release?

PR #793, opened 2026-08-05
fix(syslog): decode BSD vis escapes so CJK/UTF-8 renders correctly

### LINKED ISSUE #687: ios syslog --nojson outputs garbled CJK text as \M- escapes
## Summary

When streaming iOS device logs via `go-ios` using `ios syslog --nojson`, Chinese text from app logs is sometimes emitted as escaped sequences like `\M-f\M^H...` instead of readable UTF-8 characters.

The same log line appears correctly in Apple Console.app.

## Environment

- go-ios version: `v1.0.202`
- host OS: macOS (darwin)
- device iOS: 26.1
- command used:

```bash
ios syslog --udid "<REDACTED_UDID>" --nojson | rg --line-buffered "<REDACTED_PROCESS>\\[[0-9]+\\]"
```

## Expected behavior

Readable UTF-8 output, e.g.

```text
<Error>: ListZAHelper - checkVisibleCellsExposure - indexPath 或 model 为空 - <ZHSDShareUI.SDUICommonCell: ...>
```

## Actual behavior

Garbled/escaped text in terminal/file output, e.g.

```text
<Error>: ListZAHelper - checkVisibleCellsExposure - indexPath \M-f\M^H\M^V model \M-d\M-8\M-:\M-g\M-)\M-: - <ZHSDShareUI.SDUICommonCell: ...>
```

## Notes

- This does **not** look like app-side logging issue, because Console.app shows the same line correctly.
- Looks like an encoding/decoding path issue in `ios syslog` output pipeline.
- Reproduced with direct stdout and with pipe-to-file.

danielpaulus:
## Problem

`ios syslog --nojson` outputs garbled CJK text as `\M-` escape sequences, e.g.

```
indexPath \M-f\M^H\M^V model \M-d\M-8\M-:\M-g\M-)\M-: - <ZHSDShareUI.SDUICommonCell: ...>
```

where Console.app shows the same line correctly as `indexPath 或 model 为空 - ...`.

## Root cause

The device-side `com.apple.syslog_relay` service applies BSD `vis(3)` encoding to every byte >= 0x80 before sending log lines over the wire. Multi-byte UTF-8 sequences (CJK, emoji, accented text) therefore arrive as `\M-x` (meta) and `\M^X` (meta+control) escapes, and go-ios passed them through verbatim to stdout.

## Fix

New `syslog.DecodeVis` (`ios/syslog/vis.go`) reverses exactly the escapes syslog_relay emits for high bytes:

- `\M-x` → `0x80 | x`
- `\M^X` → `0x80 | (X ^ 0x40)`
- `\\` → `\` (needed so a literal `\M-f` in a message, wire-encoded as `\\M-f`, round-trips correctly)

It is applied in `Connection.ReadLogMessage`, i.e. before parsing/formatting, so every consumer benefits: `--nojson` raw output, legacy JSON, `--parse`, and the REST streaming endpoint.

**Decoding applies in JSON mode too — deliberately.** Go's `encoding/json` emits UTF-8 strings natively (and replaces any invalid byte sequence with U+FFFD), so decoded CJK renders as readable text inside valid JSON instead of doubly-escaped `\\M-f` noise. Keeping raw and JSON modes consistent also means downstream parsers see the same message content regardless of output mode.

**What is deliberately not decoded:** control-character escapes (`\^X`, e.g. ESC = `\^[`) and unrecognized/truncated sequences are left untouched, so decoding never reintroduces raw control bytes into terminal output (terminal-escape injection from device logs). Lone backslashes and malformed escapes pass through unchanged.

## Options considered

1. **Decode in `ReadLogMessage` (chosen)** — one seam, all output modes and the REST API fixed at once; library consumers of `ios/syslog` get readable text by default.
2. Decode only in `main.go`'s `--nojson` formatter — narrowest fix for the reported symptom, but leaves `--parse`, legacy JSON, and the REST endpoint garbled, and every consumer would have to re-implement it.
3. Full `unvis(3)` implementation (octal `\ooo`, `\^X` control chars, C-style `\n`/`\t`...) — more faithful to BSD, but syslog_relay only emits meta escapes for the bytes that matter here, and decoding `\^X` would put raw control characters (including ESC) back into terminal output. Scoped decoding is safer and sufficient.

## Test plan

- New table-driven `TestDecodeVis`: vis-escaped CJK decodes to expected UTF-8 (`\M-f\M^H\M^V` → `或`, `\M-d\M-8\M-:\M-g\M-)\M-:` → `为空`), plain-ASCII/raw-UTF-8 passthrough, escaped backslash, lone/trailing backslash, truncated `\M-`/`\M^`, unknown escapes, and control escapes staying encoded.
- New `TestParserWithDecodedVisLine`: a full syslog line with escaped CJK decodes and parses into the right Process/PID/Message.

Fixes #687

### DIFF
--- a/ios/syslog/syslog.go
+++ b/ios/syslog/syslog.go
@@ -58,12 +58,13 @@ func NewWithShimConnection(device ios.DeviceEntry) (*Connection, error) {
 
 // ReadLogMessage this is a blocking function that will return individual log messages received from syslog.
 // Call it in an endless for loop in a separate go routine.
+// Messages are vis-decoded (see DecodeVis) so non-ASCII text renders as UTF-8.
 func (sysLogConn *Connection) ReadLogMessage() (string, error) {
 	logmsg, err := sysLogConn.bufferedReader.ReadString(0)
 	if err != nil {
 		return "", err
 	}
-	return logmsg, nil
+	return DecodeVis(logmsg), nil
 }
 
 // LogEntry represents a parsed log entry
--- a/ios/syslog/syslog_test.go
+++ b/ios/syslog/syslog_test.go
@@ -59,3 +59,94 @@ func TestParser(t *testing.T) {
 		})
 	}
 }
+
+func TestDecodeVis(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "vis-escaped CJK decodes to UTF-8",
+			in:   `indexPath \M-f\M^H\M^V model \M-d\M-8\M-:\M-g\M-)\M-: - done`,
+			want: "indexPath 或 model 为空 - done",
+		},
+		{
+			name: "meta escape",
+			in:   `\M-f`,
+			want: "\xe6",
+		},
+		{
+			name: "meta-control escape",
+			in:   `\M^H`,
+			want: "\x88",
+		},
+		{
+			name: "meta-control DEL variant",
+			in:   `\M^?`,
+			want: "\xff",
+		},
+		{
+			name: "plain ASCII passes through",
+			in:   "hello world [123] <Notice>: nothing to decode",
+			want: "hello world [123] <Notice>: nothing to decode",
+		},
+		{
+			name: "empty string",
+			in:   "",
+			want: "",
+		},
+		{
+			name: "escaped backslash decodes to single backslash",
+			in:   `path C:\\Users\\M-f`,
+			want: `path C:\Users\M-f`,
+		},
+		{
+			name: "lone trailing backslash stays",
+			in:   `abc\`,
+			want: `abc\`,
+		},
+		{
+			name: "truncated meta escape stays",
+			in:   `abc\M-`,
+			want: `abc\M-`,
+		},
+		{
+			name: "truncated meta-control escape stays",
+			in:   `abc\M^`,
+			want: `abc\M^`,
+		},
+		{
+			name: "unknown escape stays",
+			in:   `abc\q def\Mx`,
+			want: `abc\q def\Mx`,
+		},
+		{
+			name: "control escape is left encoded",
+			in:   `esc \^[ bell \^G`,
+			want: `esc \^[ bell \^G`,
+		},
+		{
+			name: "raw UTF-8 passes through unchanged",
+			in:   "已经是 UTF-8 了",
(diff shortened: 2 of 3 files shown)


danielpaulus:
Adversarial review round on the vis decoder, cross-checked against Apple libc's actual `vis(3)` implementation (compiled and ran a probe against macOS libc, whose `strvisx(_, _, 0)` reproduces the exact `\M-f\M^H\M^V` output from #687) plus a Codex second opinion. Two confirmed defects, fixed in 1cc5c28:

**1. Octal escapes were not decoded (real-world mojibake).** vis octal-encodes any byte whose low 7 bits equal space, so 0xA0 — a common UTF-8 continuation byte — goes out as `\240`, not `\M- `. Empirically: `"à"` (C3 A0) encodes as `\M-C\240`, which the previous decoder turned into `Ã\240`. Also, current Apple libc emits a literal backslash as `\134` (not `\\`). `DecodeVis` now decodes `\ddd` when the value is a high byte (0x80–0xFF) or printable ASCII; control-valued octal (`\000`, `\033`, `\177`) and out-of-range (`\777`) stay encoded, consistent with the existing terminal-safety policy for `\^X`.

**2. Meta escapes accepted any payload byte.** vis only emits `\M-` + graphic char and `\M^` + `@`–`_`/`?`, but the decoder consumed any byte — e.g. `\M-` followed by a newline would eat the newline and fabricate 0x8A. Payloads are now validated; invalid sequences pass through unchanged, matching the doc contract.

**Dismissed findings:**
- "default vis emits `\040` for space" (Codex) — false; with flags=0 space passes through literally (`\040` requires `VIS_SP`). Handled harmlessly by the octal case anyway.
- "decode `\000` and all control octal per unvis(3)" (Codex) — rejected; syslog_relay never emits control bytes in octal form (they go as `\^X`), and decoding them would violate this PR's deliberate policy of not reintroducing raw control bytes into terminal output.
- `\M-\` (0xDC) and `\M^?` (0xFF) — verified correct against libc output; already handled.

Tests extended with the empirically verified encodings (`\M-C\240` → à, `\346\210\226` → 或, `\134` → `\`, boundary payloads `\M-!`/`\M-~`/`\M^@`/`\M^_`, invalid payloads, truncated octal). `go build ./...`, `go test ./...`, `gofmt -l` all clean.

PR #794, opened 2026-08-05
fix(house_arrest): fall back to VendDocuments when VendContainer is denied

### LINKED ISSUE #593: "ios fsync --udid=<udid> --app=<bundleID> tree --path=/Documents" throw error "fsync: connect afc service failed"

danielpaulus:
## Problem

`ios fsync --udid=<udid> --app=<bundleID> tree --path=/Documents` fails with:

```
{"err":"InstallationLookupFailed","level":"fatal","msg":"fsync: connect afc service failed"}
```

while `tidevice fsync -B <bundleID> ls /Documents` works on the very same devices (reported on iOS 15.7.5 and 17.6.1).

## Root cause

`ios/house_arrest` only ever sent `Command=VendContainer` to `com.apple.mobile.house_arrest`. The device denies full container access for apps that were not installed with a developer profile (i.e. App Store apps) and answers `InstallationLookupFailed`, so go-ios gave up immediately. Documents-only access via `Command=VendDocuments` is still allowed for such apps when they declare `UIFileSharingEnabled` — which is why tidevice (documents access) works on the same device.

Note: this root cause is a hypothesis backed by tidevice/pymobiledevice3 behavior on the reporters' devices; e2e confirmation on a device with an App Store app is optional (CI apps are developer-signed, so VendContainer keeps succeeding there and the fallback path stays dormant).

## Fix

- `house_arrest.New` first tries `VendContainer` as before. If the device reports an error, it reconnects (house_arrest accepts only one command per connection) and retries with `VendDocuments`, logging the fallback via `golog` with module/udid/bundleID attrs.
- If both commands fail, the returned error now includes both device errors and explains that container access needs a developer-profile install while documents access needs `UIFileSharingEnabled` in the app's Info.plist — instead of the bare `InstallationLookupFailed`.
- `checkResponse` now uses stdlib `errors` (only use of `pkg/errors` in the package) so the logged error attr isn't polluted with a stack trace.

## Options considered

1. **Automatic fallback (chosen):** `VendContainer` → on device error retry `VendDocuments`. Zero CLI changes, `--app` starts working for App Store apps exactly like tidevice, and dev-profile apps keep full container access. Paths under `/Documents` behave identically in both modes.
2. **Explicit `--documents` flag on fsync:** more explicit, but every existing caller (including `testmanagerd`'s house_arrest use) would still fail on App Store apps until users learn the flag, and it pushes a protocol detail into the CLI. Could still be added later on top of the fallback if someone needs to force documents-only mode.
3. **Both:** deferred — the fallback alone resolves the reported failure with the smallest surface.

## Test plan

- New device-free unit tests in `ios/house_arrest` with a mocked `DeviceConnectionInterface` that records the request plists and replays canned device responses:
  - `VendContainer` success → no second connection, connection kept open for AFC.
  - `VendContainer` denied (`InstallationLookupFailed`) → first connection closed, `VendDocuments` request plist emitted on a fresh connection.
  - Both denied → error mentions both device error codes and `UIFileSharingEnabled`, both connections closed.

Fixes #593

### DIFF
--- a/ios/house_arrest/house_arrest.go
+++ b/ios/house_arrest/house_arrest.go
@@ -31,6 +31,9 @@ func New(device ios.DeviceEntry, bundleID string) (*afc.Client, error) {
 // apps not installed with a developer profile, e.g. App Store apps, reporting
 // InstallationLookupFailed), it falls back to documents-only access on a fresh
 // connection, since house_arrest accepts only one command per connection.
+// Transport failures abort without the fallback: they are not a denial by the
+// device, and retrying would mask them or silently degrade a working container
+// connection to documents-only access.
 func connect(dial func() (ios.DeviceConnectionInterface, error), bundleID string, udid string) (*afc.Client, error) {
 	deviceConn, err := dial()
 	if err != nil {
@@ -41,6 +44,10 @@ func connect(dial func() (ios.DeviceConnectionInterface, error), bundleID string
 		return afc.NewFromConn(deviceConn), nil
 	}
 	deviceConn.Close()
+	var denied vendError
+	if !errors.As(containerErr, &denied) {
+		return nil, containerErr
+	}
 	golog.Info("VendContainer was denied, falling back to VendDocuments", "module", logModule, "udid", udid, "bundleID", bundleID, "err", containerErr)
 	deviceConn, err = dial()
 	if err != nil {
@@ -70,11 +77,21 @@ func vend(deviceConn ios.DeviceConnectionInterface, command string, bundleID str
 	if err != nil {
 		return err
 	}
-	return checkResponse(response)
+	return checkResponse(response, command)
 }
 
-func checkResponse(vendContainerResponseBytes []byte) error {
-	response, err := plistFromBytes(vendContainerResponseBytes)
+// vendError is an error the device reported in a house_arrest response (e.g.
+// InstallationLookupFailed), as opposed to a transport failure.
+type vendError struct {
+	message string
+}
+
+func (e vendError) Error() string {
+	return e.message
+}
+
+func checkResponse(vendResponseBytes []byte, command string) error {
+	response, err := plistFromBytes(vendResponseBytes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


danielpaulus:
## Adversarial review round (with Codex second opinion)

**Fixed (64b076b): fallback triggered on transport errors, not just device denials.** `connect` fell back to VendDocuments on *any* VendContainer failure, including send/receive/decode errors. A dropped connection was logged as "VendContainer was denied", retried, and could either mask the real transport failure behind the misleading signing/UIFileSharingEnabled error message or silently degrade a caller expecting full container access to documents-only. `checkResponse` now returns a typed `vendError` for device-reported errors, and the fallback only fires for those; transport failures return immediately with the original error (connection still closed). New unit test `TestTransportErrorDoesNotFallBack` covers this.

**Dismissed findings:**

- *"VendDocuments changes the AFC root to the Documents directory, breaking container-relative paths"* (Codex). Refuted by three data points: (1) tidevice's `fsync -B` defaults to `VendDocuments` (`_device.py: app_sync(..., command="VendDocuments")`) and the issue #593 reporter successfully ran `tidevice fsync ... ls /Documents` through it — so `Documents` is still addressable at the top of the namespace after VendDocuments; (2) pymobiledevice3's house_arrest documents VendDocuments as restricting access to "only the Documents subtree", with its shell merely *starting in* `Documents` as a client-side convenience; (3) the reported error/working-tool pair in #593 is only consistent with unchanged path semantics. `fsync --app` paths like `Documents/...` behave identically in both modes.
- *"Fallback should fire only on the exact `InstallationLookupFailed` string"* (Codex). Kept the broader device-error trigger: error strings for the denial may vary across iOS versions, and for genuinely unrelated device errors (e.g. `ApplicationLookupFailed`) VendDocuments fails identically, both errors land in the combined message, and the cost is one extra dial — nothing is masked.
- *"`deviceConn.Close()` error ignored before the variable is reused"* (Codex). Consistent with the pre-existing pattern in this file and the rest of the codebase; a failed `Close` on a device connection has no actionable recovery here.

**Verified clean:** connection lifecycle (every path either hands the conn to afc or closes it, including both-legs-fail and dial-failure paths); developer-signed apps unchanged (VendContainer succeeds on first dial, single connection, as before); `xcuitestrunner.go` house_arrest use unaffected (test runners are dev-signed, and transport errors no longer risk a silent documents-only downgrade there); logging attrs per AGENTS.md (`module` first, `udid`, `bundleID`).

PR #797, opened 2026-08-05
feat(syslog): client-side --process/--pid filters to follow the app under test

### LINKED ISSUE #202: Be able to get the MainApp logs when use ios test command
Currently, when we use the "ios test" command line we can see the logs of the test but we cannot see the log of the mainapp.

danielpaulus:
Based on #793 (stacked; same package). The diff includes #793's commit; only the last commit (`feat(syslog): add client-side --process and --pid filters`) is new here. Merge #793 first.

## Problem

Issue #202: when running XCUITests via `ios test`/`ios runtest`, there is no convenient way to see the main app's logs. `ios syslog` does stream them (NSLog/os_log go to the system log), but the full firehose makes the app's output impossible to follow.

## Root cause

`ios/syslog` already parses `Process` and `PID` out of each entry, but the CLI had no way to select entries by them — the only options were "everything" or grepping the raw stream yourself.

## Fix

- New `syslog.EntryFilter` (`ios/syslog/syslog.go`): a client-side predicate over parsed entries matching exact process name and/or PID; the zero value matches everything.
- CLI wiring: `ios syslog [--parse] [--pid=<processID>] [--process=<processName>]`. While filtering, each line is parsed and non-matching (or unparseable, hence unattributable) lines are skipped. Works with all output modes (`--nojson`, legacy JSON, `--parse`); `--pid` is validated to be numeric.
- Help text documents the runtest workflow: run `ios syslog --process=<appName>` in a second terminal while `ios test`/`ios runtest` executes to capture the app under test's NSLog/os_log output.

## Options considered

1. **Client-side filter over parsed entries (chosen)** — small, safe, works on every iOS version go-ios supports, and composes with existing output modes. Downside: the full stream still crosses USB (documented in help text).
2. Device-side filtered syslog (the "mechanism 2" mentioned in the issue thread, as in iosif) — reduces USB traffic, but is a different service/protocol and a much larger change; can still be added later without conflicting with this flag surface. For device-side filtering today, `ios ostrace --process=<name>` already exists.
3. Full Xcode-parity stdout capture inside `ios runtest` (attaching to the app process / debugger-style stdout redirection via dtx) — **out of scope**: it requires touching xctest/testmanagerd/dtx protocol internals, which are deliberately frozen until e2e coverage exists (see also the abandoned attempt in #205). `print()` output that never reaches the system log is therefore still not captured; NSLog/os_log is.

## Test plan

- New table-driven `TestEntryFilter` over canned parsed lines: empty filter matches all, match/non-match by process name (exact, not substring), match/non-match by PID, combined process+PID, nil entry never matches.
- `TestEntryFilterIsEmpty` for the zero-value fast path.
- Manual smoke test: `ios syslog --process=MyApp --pid=42` passes docopt parsing; `--pid=notanumber` is rejected.

Fixes #202

### DIFF
--- a/cmd_device_debug.go
+++ b/cmd_device_debug.go
@@ -15,6 +15,7 @@ import (
 	"github.com/danielpaulus/go-ios/ios/instruments"
 	"github.com/danielpaulus/go-ios/ios/ostrace"
 	"github.com/danielpaulus/go-ios/ios/pcap"
+	"github.com/danielpaulus/go-ios/ios/syslog"
 )
 
 func runPCAPCommand(ctx commandContext) {
@@ -35,7 +36,13 @@ func runDproxyCommand(ctx commandContext) {
 
 func runSyslogCommand(ctx commandContext) {
 	parse, _ := ctx.Args.Bool("--parse")
-	runSyslog(ctx.Device, parse)
+	process, _ := ctx.Args.String("--process")
+	pidStr, _ := ctx.Args.String("--pid")
+	if pidStr != "" {
+		_, err := strconv.Atoi(pidStr)
+		exitIfError("invalid --pid value", err)
+	}
+	runSyslog(ctx.Device, parse, syslog.EntryFilter{Process: process, PID: pidStr})
 }
 
 func runOSTraceCommand(ctx commandContext) {
--- a/cmd_device_debug.go
+++ b/cmd_device_debug.go
@@ -39,8 +39,10 @@ func runSyslogCommand(ctx commandContext) {
 	process, _ := ctx.Args.String("--process")
 	pidStr, _ := ctx.Args.String("--pid")
 	if pidStr != "" {
-		_, err := strconv.Atoi(pidStr)
+		pid, err := strconv.Atoi(pidStr)
 		exitIfError("invalid --pid value", err)
+		// normalize (e.g. "0042" -> "42") so the string compare against parsed pids works
+		pidStr = strconv.Itoa(pid)
 	}
 	runSyslog(ctx.Device, parse, syslog.EntryFilter{Process: process, PID: pidStr})
 }
--- a/ios/syslog/syslog.go
+++ b/ios/syslog/syslog.go
@@ -122,6 +122,35 @@ func Parser() func(log string) (*LogEntry, error) {
 	}
 }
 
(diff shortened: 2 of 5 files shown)


danielpaulus:
Adversarial review round (incremental diff vs #793 base) — findings fixed in 55ea128:

**Fixed**
1. **`--process` dropped most os_log lines of the target process.** Real syslog_relay lines annotate os_log senders with the emitting library: `SpringBoard(FrontBoard)[52] <Notice>: …`. The parser captures `Process = "SpringBoard(FrontBoard)"`, so the exact compare in `EntryFilter.Matches` silently filtered out exactly the lines the documented use case (capturing an app-under-test's os_log output) needs. `Matches` now also compares against the process name with a trailing `"(Library)"` annotation stripped; exact match on the full annotated name still works. Docs + 4 new table-driven test cases (including a substring-loophole guard: `MyApp` does not match `MyAppExtension(Foundation)`).
2. **`--pid` noncanonical values matched nothing.** `--pid=0042` / `--pid=+42` passed the `strconv.Atoi` validation but the raw string was compared against the parsed pid, so every line was silently dropped. The value is now normalized via `Atoi`/`Itoa` in `runSyslogCommand`.
3. **Undocumented drop of unparseable lines.** The "unparseable lines are omitted while a filter is active" behavior was only a code comment; it's now stated in the `ios syslog` help text.

**Dismissed / out of scope**
- Codex flagged that `(?P<Message>.+)` in the parser regex rejects entries with an empty message, which the filter would then drop. True, but that regex predates this PR (from #501, unchanged on `main`) and affects `--parse` identically today; changing parser semantics doesn't belong in this increment. Worth a small follow-up if empty-message lines show up in practice.
- Negative `--pid` values (e.g. `-1`) validate and match nothing — intentional: no process has that pid, and empty string is the "no filter" sentinel, so no sentinel collision exists.
- docopt wiring checked: `--pid=<processID>`/`--process=<processName>` use the same option spelling as `kill`/`pcap`/`ostrace` patterns, and dispatch via `commandByBool("syslog", …)` is unaffected. `--parse`, `--nojson`, and filter combos compose correctly (filtering happens before formatting on the vis-decoded line, consistent with what `--parse` displays).

`go build ./...`, `go test ./...`, and `gofmt -l` all clean.

PR #799, opened 2026-08-05
feat(runtest): add --junit-output flag for JUnit XML test reports

### LINKED ISSUE #286: How to access test report like xcresult or junit.xml when I run runtest
I am able to use runtest to trigger tests and am able to see the number of tests passed or failed in logs. Is there a way to access the test results like xcresult or junit xml file?

danielpaulus:
## Problem

Issue #286 asks how to get a machine-readable test report (xcresult or junit.xml) out of `ios runtest`. go-ios imitates the Xcode/testmanagerd communication and never produces an `.xcresult` bundle — that format is generated by Xcode itself, so there was no report artifact to hand to CI systems or device farms. The structured results (`[]TestSuite`) were only printed as ad-hoc JSON on stdout.

## Design

- New small pure-formatter package **`ios/junit`** that consumes the *existing* `testmanagerd.TestSuite`/`TestCase` result structs (read-only) and serializes them to standard JUnit XML: `<testsuites>` / `<testsuite>` / `<testcase>` with aggregate counts, per-case durations (seconds), suite timestamps, and `<failure>`/`<error>`/`<skipped>` children with messages and `file:line` content. XML escaping is handled by `encoding/xml`.
- Status mapping: `passed` → plain testcase; `failed` → `<failure>`; `stalled` (go-ios-specific hang detection) → `<error>`; `expected failure` → `<skipped message="expected failure...">`; any other/absent status (e.g. XCTSkip's `skipped`, or a case that never reported a result) → `<skipped>`.
- Suite `time` uses `TestDuration`; if the suite never finished regularly (duration 0) it falls back to the sum of the case durations.
- CLI: new `--junit-output=<file>` flag on `ios runtest` and `ios runxctest` (named for symmetry with the existing `--log-output=<file>`). It is additive: the JSON stdout output and `--log-output` behavior are unchanged, and the report is written even when the run errors, so partial results are preserved for CI.
- **No protocol code touched**: `ios/dtx_codec`, `ios/nskeyedarchiver`, and the testmanagerd listener/protocol code are unmodified; the formatter only reads the structs.

Known gap in the source structs: `testSuiteFinished` receives skip/expectedFailure counters from the device but the `TestSuite` struct doesn't retain them, so suite-level counts are derived from the per-case statuses instead (which is what JUnit consumers care about anyway).

## Implementation

- `ios/junit/junit.go`: `junit.Write(w io.Writer, suites []testmanagerd.TestSuite) error` — converts and encodes with the XML header and indentation.
- `cmd_device_testing.go`: `writeJUnitReport` helper; `runTestCommand` and `runXCTestCommand` now build the listener in the branch and share a single run/output path, then write the JUnit file when `--junit-output` is set.
- `main.go`: docopt usage lines + help text for `runtest`/`runxctest`.

## Options considered

1. **Generate `.xcresult`** — rejected: xcresult is an opaque, Xcode-generated bundle format; reproducing it outside Xcode is impractical and unsupported. JUnit XML is the lingua franca CI systems (Jenkins, GitLab, Buildkite, device farms) actually ingest, which is the underlying need in #286.
2. **Emit JUnit from inside `testmanagerd`** — rejected: keeps report formatting out of the protocol layer; the CLI-side formatter consumes the public result structs, so library users can call `junit.Write` too.
3. **Generic `--output=<path>` flag** — rejected in favor of the explicit `--junit-output=<file>` (preferred): it mirrors the existing `--log-output=<file>` convention, avoids colliding with the `--output=<outfile>` meaning used by screenshot/wallpaper commands, and leaves room for other formats later.

**Preferred: option 3's explicit flag + option 2's separate `ios/junit` package** — smallest, additive change that serves the CI/device-farm use case without touching protocol code.

## Test plan

- Unit tests in `ios/junit/junit_test.go`: fixture `TestSuite` slices → exact expected XML for pass/fail/skip/stalled/expected-failure, durations and duration fallback, timestamps, failure messages with `file:line`, XML escaping of `<`, `>`, `&`, quotes, multi-suite aggregate counts, and the empty report.
- `go build ./...` and full `go test ./...` pass; `gofmt` clean.
- Real xcresult bundles are Xcode-generated and out of scope; JUnit XML serves the CI/device-farm use case. Real-device verification via the e2e suite (`/test-devices`).

Fixes #286


danielpaulus:
Adversarial review round (Claude + Codex second opinion) — 1 fix pushed (d101593).

**Fixed**
- **Never-finished test cases reported as `<skipped>`** (`ios/junit/junit.go`): a `TestCase` with empty `Status` means it started but never got a final result from testmanagerd — app crash, lost connection, or interrupted run (`testCaseDidFinishForTest` never fired). Emitting `<skipped>` made such a run look green in CI. Now emits `<error message="no test result received">`, matching JUnit error semantics ("an unanticipated problem prevented the test from completing"). Stalled tests already mapped to `<error>`; XCTSkip-style statuses still map to `<skipped>`. Tests updated.

**Checked and found correct**
- Invalid XML characters (control chars in failure messages): Go's `encoding/xml` encoder replaces XML-forbidden runes with U+FFFD in both attributes and chardata — output stays well-formed, verified experimentally. Existing escaping test covers `<`, `&`, quotes.
- JUnit shape: `testsuites`/`testsuite`/`testcase` attrs (`tests`, `failures`, `errors`, `skipped`, `time`, `timestamp`) are the shape Jenkins/GitLab/Buildkite consume; `time` as fixed 3-decimal seconds and zone-less ISO8601 `timestamp` are the conventional formats.
- Refactored `runtest`/`runxctest` path preserves prior behavior exactly: same listener wiring per `--log-output`, same JSON on stdout, same log-and-continue on run error, same exit codes.
- Partial-report-on-error: `RunTestWithConfig` returns collected suites alongside the error (`xcuitestrunner.go:445`), and the JUnit file is written after the JSON print, so partial results land in the report.

**Dismissed (with reasons)**
- Codex: "run errors only logged, process exits 0, so CI can go green on infrastructure failure" — pre-existing CLI behavior on main; this PR intentionally preserves exit-code semantics. Worth a separate discussion (e.g. a `--fail-on-error` flag or non-zero exit on failed tests), not a defect introduced here. The green-on-crash half of that finding is addressed by the fix above.
- Codex: "expected failure should not map to `<skipped>`" — JUnit has no expected-failure concept; mapping to `<skipped message="expected failure: …">` follows the same convention pytest's junitxml uses for xfail, and doesn't fail builds. Defensible as-is.

Issue #784, opened 2026-07-23
iOS 26.5.x: file operations on --app-group containers fail device-side with RemoteServices 11007 "File paths cannot contain '..'" (paths contain no '..'); file ls hangs silently; failed pull leaves zero-byte local file


bitjson:
Root cause is an Apple issue, but dropping this here in case the go-ios suggestions are useful. 

(Fable 5:)

> ### Summary
> 
> On devices running iOS 26.5.2, `ios file pull --app-group=<groupID>` fails for app-created files in the App Group container with a device-side error:
> 
> > code 11007, `File paths cannot contain '..'.`
> 
> although neither the submitted relative path nor the device-resolved absolute path contains `..`. `ios file ls --app-group=…` (root or any subdirectory) hangs indefinitely with no output. The same operations against the app's own data container (`--app=<bundleID>`) work fine on the same devices and tunnels.
> 
> **This does not appear to be a go-ios framing bug** — I verified that Apple's own `xcrun devicectl` (518.33, Xcode 26.6, macOS 26.5.2 host, freshly paired) fails identically on the same paths, and go-ios sends the caller's path verbatim (`ios/fileservice/fileservice.go:178-182`). The rejection happens on the device after correct container resolution: the device's encoded error names the fully-resolved absolute container path, and its UUID matches what the app itself reports from `containerURL(forSecurityApplicationGroupIdentifier:)`. Reproduced identically on three devices (iPhone 15 Pro Max, iPhone 16 Pro, iPhone 16 Pro Max, all iOS 26.5.2), across two independent client stacks (go-ios's own tunnel/RemoteXPC and Apple's CoreDevice). Filing here because `--app-group` is an advertised go-ios capability that is currently unusable on iOS 26.5.x, and because two go-ios-side behaviors around the failure could be hardened (see "Suggested improvements").
> 
> ### Reproduction
> 
> Requires: a macOS or Linux host with go-ios v1.2.0 (also current `main` @ 274bc43 — the fileservice code is unchanged since PR #642), an iOS 26.5.x device in developer mode, and any dev app entitled with an App Group.
> 
> 1. In the app, create nested content in the App Group container and print the root:
> 
>    ```swift
>    let fm = FileManager.default
>    let root = fm.containerURL(forSecurityApplicationGroupIdentifier: "group.example.sync")!
>    let dir = root.appendingPathComponent("shared-data/sub/dataset-a", isDirectory: true)
[...]

PR #802, opened 2026-08-05
fix(fileservice): don't hang on ListDirectory failures; don't leave zero-byte files on failed pulls (iOS 26.5.x app-group, #784)

danielpaulus:
## Problem

On iOS 26.5.x, file operations against App Group containers (`--app-group`) fail device-side with RemoteServices error 11007 "File paths cannot contain '..'." even though the paths contain no `..` (#784). That is an Apple daemon bug — reproduced identically with Apple's own `xcrun devicectl` on the same devices — but go-ios failed badly around it:

- `ios file ls --app-group=…` **hung forever** with no output.
- A failed `ios file pull` left a **zero-byte local file** behind, which scripted pipelines can mistake for a successfully pulled empty file.

## Root cause

- `ListDirectory` sends `RetrieveDirectoryList` and then blocks in `ReceiveOnClientServerStream()` with no timeout. Every *other* control response (CreateSession, RetrieveFile, ProposeFile) arrives on the **server→client** stream — when the device reports a `RetrieveDirectoryList` failure there instead of sending the listing, the error was silently lost and the call blocked indefinitely.
- `cmd_device_files.go` ran `os.Create(localPath)` *before* `PullFile`, so any pull failure (including the device-side 11007 rejection, which happens before any data streaming) left a truncated 0-byte destination file.

## Fix

- **`ios/fileservice`**: `ListDirectory` now receives on both streams concurrently under a receive timeout (default 30s) and surfaces failures as typed errors: `*fileservice.DeviceError` (carries the device's `EncodedError` + `LocalizedDescription`) and `fileservice.ErrTimeout`. A pending control-stream read left by a successful listing is tracked on the `Connection` and consumed by the next control receive (`createSession`/`PullFile`/`PushFile`), so the control stream never has two concurrent readers. `extractError` now returns `*DeviceError` everywhere (same message format as before).
- **`cmd_device_files.go`**: the pull path is extracted into a testable `pullToLocalFile` wrapper that removes the local file when the download fails — no zero-byte/partial files remain. Success behavior (including pulling genuinely empty files) is unchanged.
- **Interface seam**: `Connection.conn` is now a minimal `controlConnection` interface (satisfied by `*xpc.Connection`) so the control protocol is unit-testable without a device. No change to the wire protocol or `ios/xpc`.
- **Docs**: `ios file ls` help text in `main.go` now notes the iOS 26.5.x App Group limitation and the `--app` workaround.

## Options considered

1. **Timeout-only on the list stream, check control stream after expiry** — simplest, but the device's 11007 error would only surface after the full timeout instead of immediately, and the error could still be lost. Rejected.
2. **Concurrent receive on both streams with a pending-read handoff (chosen)** — surfaces the device error immediately, keeps a single reader per stream, and reduces to the old behavior on the happy path. Slightly more state (`pendingControl`), fully covered by tests.
3. **Modify `ios/xpc` to support deadlines/contexts** — cleaner long-term, but a much larger blast radius across all XPC consumers; out of scope for a targeted hardening.
4. For the pull: **create the local file only after the control response** (needs restructuring `PullFile`'s writer contract and breaks empty-file pulls) vs. **unlink on error (chosen)** — minimal, and also cleans up partially-written files on mid-stream failures.

## Test plan

- New device-free unit tests that hang (and fail via a 5s test deadline) without the fix:
  - `ios/fileservice/fileservice_test.go` against a fake XPC connection: control-stream `EncodedError` → typed `*DeviceError`, no hang; device never responds → `ErrTimeout` with short injected timeout; happy-path listing; pending control-read handoff to `PullFile`.
  - `cmd_device_files_test.go`: erroring puller (immediate and mid-stream) → no local file remains; successful pull writes content and returns size.
- Verified the ListDirectory tests fail against the pre-fix receive logic (both hang), and pass with the fix.
- `go build ./...`, `go test ./...`, `go test -race` on changed packages, `gofmt -l` clean.

Improves #784 — the go-ios-side hang and zero-byte-file behaviors are fixed here, but the underlying App Group 11007 rejection is an Apple iOS 26.5.x daemon bug (Apple's devicectl fails identically), so the issue stays open to track that limitation.

### DIFF
--- a/cmd_device_files.go
+++ b/cmd_device_files.go
@@ -170,29 +170,40 @@ func runFileCommand(ctx commandContext) {
 }
 
 // pullToLocalFile downloads remotePath into localPath using the passed pull
-// function and returns the downloaded size. If the pull fails, the local file
-// is removed so a failed pull does not leave a zero-byte or partial file
-// behind that pipelines could mistake for a successful download (issue #784).
-func pullToLocalFile(pull func(remotePath string, writer io.Writer) error, remotePath, localPath string) (int64, error) {
+// function and returns the downloaded size. On any failure the local file is
+// removed so a failed pull never leaves a zero-byte or partial file behind that
+// pipelines could mistake for a successful download (issue #784). This includes
+// failures surfaced by Close, which is where buffered write errors such as a
+// full disk are reported.
+func pullToLocalFile(pull func(remotePath string, writer io.Writer) error, remotePath, localPath string) (size int64, err error) {
 	outputFile, err := os.Create(localPath)
 	if err != nil {
 		return 0, fmt.Errorf("failed to create output file: %w", err)
 	}
+	// Remove the file on any error path below. os.Create truncated any prior
+	// contents, so there is nothing worth keeping if the pull does not complete.
+	defer func() {
+		if err != nil {
+			os.Remove(localPath)
+		}
+	}()
 
-	if err := pull(remotePath, outputFile); err != nil {
+	if pullErr := pull(remotePath, outputFile); pullErr != nil {
 		outputFile.Close()
-		os.Remove(localPath)
+		err = pullErr
 		return 0, err
 	}
 
-	fileInfo, err := outputFile.Stat()
-	if err != nil {
+	fileInfo, statErr := outputFile.Stat()
+	if statErr != nil {
 		outputFile.Close()
-		return 0, fmt.Errorf("failed to get file info: %w", err)
+		err = fmt.Errorf("failed to get file info: %w", statErr)
+		return 0, err
 	}
 
-	if err := outputFile.Close(); err != nil {
-		return 0, fmt.Errorf("failed to close output file: %w", err)
+	if closeErr := outputFile.Close(); closeErr != nil {
+		err = fmt.Errorf("failed to close output file: %w", closeErr)
+		return 0, err
 	}
 
 	return fileInfo.Size(), nil
--- a/cmd_device_files.go
+++ b/cmd_device_files.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"path"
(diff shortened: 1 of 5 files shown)


danielpaulus:
Adversarial review (with a second-opinion pass via codex). Verdict: the core design is sound; I pushed one follow-up commit for a real cleanup gap and dismissed the rest as false positives or documented-by-contract.

## Fixed (commit `932e266`)
**`pullToLocalFile` left a partial file behind on Stat/Close failure.** The wrapper documents that a failed pull never leaves a zero-byte/partial file, but it only removed the file when the `pull` callback errored. If `pull` succeeded yet `Stat` or `Close` failed, it returned an error while leaving the incomplete file on disk. `Close` is precisely where buffered write errors (e.g. a full disk) surface, so this was the case most likely to leave a truncated file a pipeline mistakes for success. Cleanup now runs via a deferred, error-guarded `os.Remove` on every post-`os.Create` failure path, plus a regression test for the finalize-failure path.

## Reviewed and dismissed
- **Goroutine leak / protocol desync when the control stream wins or on timeout.** Real in the abstract, but the losing goroutine only leaks until `Close()`, and both `ErrTimeout` and the `ListDirectory` doc mandate closing the connection after an error/timeout. The sole caller (`cmd_device_files.go`) creates a fresh connection per CLI invocation and always `defer`s `Close()`, so the reuse-after-error desync never materializes. Not a bug given the documented contract.
- **Data race on `pendingControl` / `conn`.** No race: all `pendingControl` access is under `c.mu`; `conn` is set once in `New` and never reassigned (only `Close()`d). `go test -race ./ios/fileservice/` is clean.
- **TOCTOU on `os.Remove` / truncating a pre-existing destination.** The "another process swaps the path between Close and Remove" scenario is not a realistic threat for a user-specified CLI download target, and `os.Create` truncating an existing destination matches the prior behavior on `main` and every download tool. No change warranted.

## Test status
`go build ./...`, `go test ./...`, and `go test -race ./ios/fileservice/` all pass; `gofmt -l` clean.

danielpaulus:
Thanks for tackling the #784 fileservice hang! Heads-up from a CI triage pass: the real-device e2e is failing here for a reason that looks like it needs another look at this change, not an infra flake.

`TestFileLsCrash` and `TestFileLsTemp` fail **identically on both the macOS and Linux runners** (two different healthy iOS 17 devices — the `file-roundtrip` push/pull subtest passes on both, so the devices are fine), with:

```
ListDirectory: no response within 30s: timed out waiting for a response from the device
```

That 30s timeout is the new code path this PR introduces in `ios/fileservice/fileservice.go` — the concurrent list-stream + control-stream reader racing the `receiveTimeout`. It looks like the **normal (success) path** of `ListDirectory` now regresses: neither the list channel nor the control channel fires, so it always hits the timeout branch. The most likely culprit is the concurrent control-stream reader (or a cached `pendingControl` read) consuming/stealing the directory-list response the main read is waiting for.

So the fix for the hang-on-failure case seems to have introduced a hang on the happy path. Could you take another look at the `ListDirectory` rewrite so a normal successful listing still returns? Happy to help verify on the device farm once updated. 🙏

PR #803, opened 2026-08-05
fix(imagemounter): resolve developer disk images offline before any download

### LINKED ISSUE #644: Offline method mount images to device
@danielpaulus 

can we have a functionality to mount the images to iOS devices in offline mode or secure network

as *ios image auto*  command first download the image from internet and then it mount to connected device,.
response would be greatly appreciated .

danielpaulus:
## Problem

`ios image auto --basedir=...` is supposed to reuse images that are already present in the basedir, but users with pre-downloaded images still saw it hit the network — and fail entirely when offline or on a locked-down network (#644, reproduced by multiple reporters with images already in `devimages`).

## Root cause

Two independent cache-lookup gaps in `ios/imagemounter/imagedownloader.go` caused a download despite a present image:

1. **iOS 17+ (`Download17Plus`)**: the basedir lookup only matched the *exact pinned* DDI directory name (`ddi-17E5179g`). Any personalized DDI already present under a different name — e.g. `ddi-15F31d` downloaded by an older go-ios before the pin was bumped, or a DDI placed manually — was ignored, and a pre-downloaded `ddi-*.zip` archive was never considered. Result: a fresh download from deviceboxhq every time, which fails offline.
2. **iOS <17 (`DownloadImageFor`)**: for catalog versions whose name carries a build id (`8.4 (12H141)` … `12.3 (16F148)`), the lookup searched `<version> (<build>)/DeveloperDiskImage.dmg` while the download itself writes into the short `<version>/` directory. go-ios could never find its own previous download; the retry then failed on `os.Mkdir` (`file exists`) or re-downloaded.

## Fix

The basedir is now always consulted before any HTTP request:

- **iOS 17+**: `findPersonalizedDDI` recognizes a personalized DDI by *layout* (`Restore/BuildManifest.plist` — exactly what `PersonalizedDeveloperDiskImageMounter.MountImage` needs) instead of by directory name, preferring the pinned name when several DDIs are present. A present `ddi-17E5179g.zip` is extracted locally instead of downloaded (falling back to download if the archive is corrupt).
- **iOS <17**: the lookup checks both the full `<version> (<build>)` layout (manual downloads mirroring the upstream mspvirajpatel repo) and the short `<version>` layout that go-ios writes; `os.Mkdir` became `os.MkdirAll` so a leftover version dir no longer aborts a legitimate download.
- `DownloadImageFor` now delegates to a device-free `downloadImageForVersion(baseDir, productVersion, udid)` helper so the resolution logic is unit-testable; the public API is unchanged (`restapi` and the CLI keep calling `DownloadImageFor`). `devicebox` became a package var so tests can point it at a local server.

## Options considered

1. **Structure-based DDI detection + dual-layout lookup (chosen)** — finds any usable image that is genuinely present, regardless of how/when it got there; no API or CLI changes; download behavior on a true miss is unchanged.
2. Only widen the pinned-name check to `ddi-*` glob — would still miss manually placed DDIs (e.g. extracted from Xcode or other tools) and wouldn't fix the iOS <17 short/full directory mismatch.
3. Add an explicit `--offline` flag that forbids downloads — doesn't fix the underlying cache misses (a present image would still be ignored), and reporters' expectation is that presence in the basedir is enough. Could still be added later on top.

Preferred: option 1 — it makes "image is in the basedir" sufficient, which is what the issue asks for, with the smallest surface change.

## Test plan

- New device-free unit tests in `imagedownloader_internal_test.go` pre-populate a temp basedir and point *all* download URLs (devicebox + versionMap) at an `httptest` server whose handler fails the test on any request:
  - iOS 17+: DDI present under a non-pinned name; pinned name preferred when both exist; pre-placed zip archive extracted offline.
  - iOS <17: image found in the short `12.3/` layout (go-ios' own download layout — failed before this change) and in the full `12.3 (16F148)/` layout; plus a real-miss test asserting download into the short dir and offline reuse on the second resolution.
- `go build ./...`, `go test ./...`, and `gofmt -l` clean.
- Real-device e2e suite via CI.

Fixes #644

### DIFF
--- a/ios/imagemounter/imagedownloader.go
+++ b/ios/imagemounter/imagedownloader.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"time"
 
@@ -91,11 +90,17 @@ var (
 const (
 	imageFile     = "DeveloperDiskImage.dmg"
 	signatureFile = "DeveloperDiskImage.dmg.signature"
-	devicebox     = "https://deviceboxhq.com/"
 	// iOS 17+ universal personalized developer disk image hosted on deviceboxhq.
 	// Bump this when a newer DDI is published there (was ddi-15F31d).
 	xcode15_4_ddi = "ddi-17E5179g"
+	// buildManifestFile identifies the personalized developer disk image
+	// layout (iOS 17+): a 'Restore' directory containing a BuildManifest.plist.
+	buildManifestFile = "BuildManifest.plist"
 )
+
+// devicebox hosts the personalized developer disk image. It is a var so unit
+// tests can point it at a local test server.
+var devicebox = "https://deviceboxhq.com/"
 
 func MatchAvailable(version string) string {
 	golog.Debug("matching available image for device version", "module", logModule, "version", version)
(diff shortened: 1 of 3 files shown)


danielpaulus:
## Adversarial review — findings + one fix pushed

Reviewed the offline-cache + path-containment commits deeply (manual trace + Codex read-only pass, claims verified against the code and threat model).

### Fixed (pushed as 99ae2fe)
**Non-deterministic / false-positive DDI selection in `findPersonalizedDDI`.** The pinned-DDI preference used `strings.Contains(dir, xcode15_4_ddi)` over the *whole Restore path* and otherwise took the first walked entry. A decoy directory whose name merely contains the pinned string (e.g. `aaaa-ddi-17E5179g-copy`) could out-rank the directory named exactly like the pinned DDI, and a pinned string in an unrelated *ancestor* path could hijack the choice. The failure mode is bad: it returns a bogus `Restore` dir and never falls back to downloading, so the mount fails with a confusing error.

Fix: the preference now keys off the directory that holds `Restore` (what go-ios names after the DDI), in a deterministic order — exact pinned name, then name containing the pinned string, then the lexicographically-first structurally-valid image. Added two regression tests (`TestDownload17PlusPrefersExactPinnedDDIName` is decisive: it fails on the pre-fix code, picking the decoy, and passes after).

### Dismissed (with reasons)
- **`safeJoin` follows symlinks planted inside basedir (Codex High).** True lexically, but not a practical new bypass: planting a symlink inside basedir already requires write access to basedir, at which point an attacker can write there directly. `basedir` is operator-controlled and `safeJoin`'s contract is `..`/absolute-element containment, which it meets. Out of scope for this PR's threat model.
- **REST `basedir` accepts absolute paths (Codex High).** Intentional — absolute `--basedir` is a documented, supported feature (the e2e suite passes absolute temp dirs). The `..` check only stops upward traversal relative to a chosen base; blocking absolute paths would be a regression. The endpoint is device-auth'd.
- **First-run / empty-basedir regression.** None — `MkdirAll` creates the tree and both iOS 17+ and pre-17 cache-miss paths download exactly as before. `ios.Unzip` already has zip-slip protection (covered by the new `TestDownload17PlusRejectsZipSlipArchive`). Corrupt-zip fallback to download is tested.

### Status
`go build ./...` clean, `go test ./...` green, `gofmt` clean. CodeQL: all three Analyze jobs + CodeQL check **pass**. The image-path e2e tests (`TestImageList`, `TestImageAuto`, `TestImageMount`) pass on both runners. The Linux e2e job failed only on unrelated `usbmuxd: broken pipe` / tunnel flakes (`TestPs`/`TestScreenshot`/`TestLaunchKill`); the same suite is green on macOS.

PR #804, opened 2026-08-05
fix(forward): reconnect automatically after device replug (ios forward)

### LINKED ISSUE #378: ios forward not reconnecting automatically if device disconnected and reconnected
**Describe the bug**
ios forward command not reconnecting when device is disconnected and connected back.

**To Reproduce**
1. Connect iOS device
2. Forward port using command "ios forwad 7777 9100"
3. Load the localhost - http://localhost:7777 - Works
4. Disconnect device
5. Connect device again
6. Load the localhost - http://localhost:7777  - Not connecting

**Expected behavior**
When reconnecting the device , the forwarded port should work automatically.
The same steps works with iproxy.exe which is from libimobiledevice.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: Windows 10 Enterprise 22H2 19045.4046

**Smartphone (please complete the following information):**
 - Device: iPhone 12
 - OS: 16.5.1

danielpaulus:
## Problem

`ios forward <hostPort> <phonePort>` stops working after the device is unplugged and plugged back in. Every new client connection fails with `could not connect to phone` until the forward process is restarted. `iproxy` from libimobiledevice handles a replug in the same session.

## Root cause

`Forward()` captures the numeric usbmux `DeviceID` once at startup and every proxied connection dials usbmuxd with it (`forward.go`, previously line 95-97). usbmuxd assigns a **new** device id each time a device is plugged in, so after a replug the captured id is stale and every `Connect` request is answered with a mux error. The device's udid (serial) is stable across replugs — only the numeric id changes.

## Fix

- Each accepted client connection still dials with the captured id first (zero extra cost on the happy path).
- When that connect fails, the device is re-resolved by udid via an injected `resolveDeviceID` func (backed by `ios.GetDevice(udid)`), and the connect is retried once with the fresh id.
- The usbmux socket is now closed when `Connect` fails (previously it leaked on that path).
- The exported `StartNewProxyConnection` keeps its signature and behavior (no resolver); the resolver is wired in by `Forward()`.

## Options considered

1. **Re-resolve by udid on connect failure, retry once (chosen).** Happy path unchanged, one extra usbmuxd round trip only while the id is stale, no background goroutines or listeners.
2. **Re-resolve on every accepted connection.** Simpler control flow but adds a `ListDevices` round trip to every connection even when nothing changed.
3. **Subscribe to usbmuxd attach/detach events and update the id.** Most "live", but adds a long-lived listener connection and state shared across goroutines for a case the retry handles with less machinery.

## Test plan

Device-free unit tests (`ios/forward/forward_reconnect_test.go`) run a minimal fake usbmuxd on a temp unix socket (`USBMUXD_SOCKET_ADDRESS`) that rejects a stale id with mux error 2 and echoes bytes for the current id:

- `TestStaleDeviceIDIsReResolved` — reproduces #378: stale id triggers exactly one re-resolve, and bytes flow end-to-end afterwards. Does not compile/pass without this change.
- `TestCurrentDeviceIDConnectsWithoutResolve` — happy path never invokes the resolver.
- `TestStaleDeviceIDWithoutResolverFails` — pins the exported `StartNewProxyConnection` behavior: stale id fails and the client conn is closed.

`go build ./...`, `go test -race ./...` and `gofmt -l` are clean.

Fixes #378

### DIFF
--- a/ios/forward/forward.go
+++ b/ios/forward/forward.go
@@ -24,6 +24,13 @@ type ConnListener struct {
 	quit     chan interface{}
 }
 
+// resolveDeviceID returns the current usbmux device id of the forwarded device.
+// usbmuxd assigns a fresh id every time a device is plugged in, so the id
+// captured when the forward started goes stale after a disconnect/reconnect.
+// Re-resolving through this func lets a long-running forward survive a replug
+// (issue #378).
+type resolveDeviceID func() (int, error)
+
 // Forward forwards every connection made to the hostPort to whatever service runs inside an app on the device on phonePort.
 // Port values must be between 1 and 65535.
 func Forward(device ios.DeviceEntry, hostPort uint16, phonePort uint16) (*ConnListener, error) {
@@ -43,9 +50,28 @@ func Forward(device ios.DeviceEntry, hostPort uint16, phonePort uint16) (*ConnLi
 		quit:     make(chan interface{}),
 	}
 
-	go connectionAccept(cl, device.DeviceID, phonePort)
+	go connectionAccept(cl, device.DeviceID, phonePort, deviceIDResolver(device))
 
 	return cl, nil
+}
+
+// deviceIDResolver builds a resolveDeviceID that looks the device up again by
+// its udid (the serial survives a replug, the usbmux id does not). Without a
+// udid there is nothing stable to search by, so the captured id is returned
+// as-is.
+func deviceIDResolver(device ios.DeviceEntry) resolveDeviceID {
+	udid := device.Properties.SerialNumber
+	deviceID := device.DeviceID
+	if udid == "" {
+		return func() (int, error) { return deviceID, nil }
+	}
+	return func() (int, error) {
+		fresh, err := ios.GetDevice(udid)
+		if err != nil {
+			return 0, err
+		}
+		return fresh.DeviceID, nil
+	}
 }
 
 // Close stops listening on the host port for the forwarded connection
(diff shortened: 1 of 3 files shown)


danielpaulus:
Adversarial review (with codex cross-check). Verdict: **ship-worthy**, one small test-hygiene fix pushed to this branch.

**Fix applied** (`85e4fbb`)
- `forward_reconnect_test.go`: `startFakeUsbmuxd` did `os.MkdirTemp` then `net.Listen`, but registered the temp-dir cleanup only *after* the `net.Listen` check. A `net.Listen` failure `t.Fatalf`s before cleanup is set up, leaking the dir. Moved `t.Cleanup(os.RemoveAll)` immediately after `MkdirTemp` and split the listener close into its own cleanup so ordering (listener closed before dir removed) is preserved.

**Reviewed, no change needed**
- usbmux socket lifecycle: the real leak is fixed — `dialPhonePort` now calls `usbmuxConn.Close()` on `Connect` failure, which the old code never did. The `NewUsbMuxConnectionSimple` error path correctly does not Close (deviceConn is nil there).
- resolver-error / stale-id-without-resolver paths close the client conn and are covered by `TestStaleDeviceIDWithoutResolverFails`.
- `USBMUXD_SOCKET_ADDRESS` test hygiene: uses `t.Setenv` (auto-restored, blocks `t.Parallel`), no cross-test bleed.

**Dismissed** (codex flagged, not a bug)
- "same device id ⇒ no retry violates retry-once": intentional. #378 is specifically a *changed* usbmux id on replug; retrying the identical id against deterministic usbmuxd would just fail again and double the connect load on every transient failure. Reconnecting only when the id actually changed is the correct scope.

Verified: `go build ./...`, `go test -race ./ios/forward/`, `gofmt -l` all clean.

PR #809, opened 2026-08-05
fix(tunnel): TunnelManager self-heals dead tunnels, stops retrying unsupported devices, tested udid filter

### LINKED ISSUE #479: Is it possible to remove the output of the tunnel start command?
I am using it on Windows OS.

I am using it well with the ios tunnel start command.

However, I want to hide the output when reconnecting the device with the output of iOS 17 or lower. Is there a way to do this?

{"error":"manualPairingTunnelStart: unsupported iOS version 15.8.3","level":"warning","msg":"failed to start tunnel","time":"2024-09-20T17:04:09+09:00","udid":"xxxxx"}
{"level":"info","msg":"event: 0","time":"2024-09-20T17:04:52+09:00"}

### LINKED ISSUE #523: Using ios <17 and >=17 together is generating a huge amount of logging
**Describe the bug**
While having an iPhone with iOS 16 or 15 connected simultaneously with an iOS 17 or 18, the tunnel process output has a message every second saying it can't start the tunnel for the 16. This message escalates when adding more older devices connected with the 17 and 18 ones.

```
{"error":"manualPairingTunnelStart: unsupported iOS version 16.6.0","level":"warning","msg":"failed to start tunnel","time":"2024-12-03T10:09:31-08:00","udid":"00008020-000264E9026A002E"}
``` 

**To Reproduce**
1. Connect both iOS 17/18 and 15/16 devices to the same machine
2. Start the tunnel with ios tunnel start

**Expected behavior**
Tunnel output messages do not contain error messages that are not an error, just a regular usage condition.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: any OS

**Smartphone (please complete the following information):**
 - Device: iPhone XR (ios 16), 13 (ios 17) and 14 (ios 18)

### LINKED ISSUE #607: How to tunnel start for multiple iOS devices?
I connected two iOS devices to my computer and attempted to run the following command：

> ios tunnel start --udid=00008120-000448862178A01E

and then ：> ios tunnel ls

Image

I see The udid parameter seems to be ineffective, as it only creates one tunnel for another phone

Does anyone know how can I create two tunnels so that I can operate two iPhones at the same time ? OR let me tunnel start one by one without Manually disconnect a device?

### LINKED ISSUE #765: TunnelManager agent doesn't auto-recover a dead-but-still-present tunnel (reboot test only covers explicit refresh)
## Summary

The running tunnel agent (`ios tunnel start`) never automatically rebuilds a tunnel whose underlying transport has died **while the device is still present on USB** — e.g. after a quick device reboot that usbmux doesn't observe as a disconnect, or a tunnel transport death with no USB drop. The agent keeps serving the stale RSD record forever, and every consumer dialing it fails (see #764 for the 135s-no-timeout amplification).

`TestTunnelRebootRecovery` exists but does **not** guard this path: it recovers by **explicitly calling `ios tunnel refresh`**, which the production agent never does. So the on-demand command is tested; the agent's automatic self-heal is not.

## Live evidence (real CI host)

On the Linux tunnel runner, device `00008110-001C19A43420401E` (iPhone14,2, iOS 18.4.1):

- usbmux lists the device fine — it never left the device list.
- The agent still serves the **stale** record `{"address":"fdb2:7531:10d5::1","rsdPort":51971,"userspaceTun":false}` — the exact dead endpoint e2e tests timed out on across two separate runs.
- Because the record persisted unchanged, the agent's cleanup (which only stops tunnels for devices that *vanish* from usbmux) never fired, and `UpdateTunnels` skips any device that already has a record (`if _, exists := localTunnels[udid]; exists { continue }`) — so it was never rebuilt.

## Root cause

`TunnelManager.UpdateTunnels`:
- **creates** a tunnel only for a device with **no** existing record, and
- **tears down** a record only when the device **disappears** from `ListDevices()`.

There is **no liveness/health check of an existing tunnel record**. A tunnel that dies without a corresponding usbmux disconnect becomes a permanent stale entry. #738 added reboot recovery, but only for the disconnect-driven path (device leaves USB → cleanup → rebuild) and the on-demand `tunnel refresh` command.

`TestTunnelRebootRecovery` even acknowledges the gap and works around it:
```go
} else {
    t.Logf("did not observe %s leave USB (it may have rebooted quickly); proceeding to recovery", udid)
}
// ... then explicitly:
harness.TryRun(t, "tunnel", "refresh", "--udid="+udid, portArg)
```

## Proposed fix

1. **Agent-side auto health-check + refresh.** In the `UpdateTunnels` loop, periodically probe existing tunnel records (cheap RSD/RemoteXPC handshake or a short-timeout dial). On failure, treat the record as stale → tear it down and rebuild for that device (reuse #738's per-device refresh). This makes the agent do automatically what the test does by hand.
2. **Pair with the fast-fail dial timeout from #764** so a dead tunnel is detected in seconds, not the ~135s kernel default.
3. **Strengthen the test.** Extend `TestTunnelRebootRecovery` (or add a sibling) to assert the agent recovers a rebooted device's tunnel **without** an explicit `tunnel refresh` call — i.e. the running agent self-heals on its own. Today the test passing gives false confidence.

## Related

- #764 — kernel-tunnel dead-endpoint dials block ~135s (no dial timeout); same incident, complementary fix (fast-fail + staleness signal).
- #738 — added per-device tunnel stop/refresh and the disconnect-driven recovery this issue extends.

danielpaulus:
**Stacked PR — based on #808.** This branch includes #808's commit (the shared dial-timeout helper its liveness probe reuses) and should merge **after** #808; once #808 lands, this diff reduces to the TunnelManager changes only.

## Problem

Three long-standing tunnel-agent problems, all in `TunnelManager.UpdateTunnels`:

1. **#765 — no self-heal for dead-but-still-present tunnels.** A tunnel that dies without a usbmux disconnect (quick reboot the mux never observes, transport death while the TUN route stays up) remains a stale record forever: the create path skips devices that already have a record, and teardown only fires for devices that vanish from `ListDevices()`. Every consumer then dials the dead endpoint (135s each before #808/#764). Observed live on the Linux CI runner across two runs.
2. **#523 — sub-iOS17 devices spam warnings forever.** With mixed iOS <17 / 17+ fleets, every update cycle logs `failed to start tunnel ... unsupported iOS version` for each old device, escalating with device count.
3. **#607 / #479 — per-device tunnel management.** Users asked for `ios tunnel start --udid` to manage a single device's tunnel. The `TunnelManager.udidFilter` plumbing exists (`NewTunnelManagerForDevice`, wired from `cmd_tunnel.go` with help text in `main.go`) but had zero unit-test coverage guarding it.

## Root cause

`UpdateTunnels` had exactly two behaviors: create-if-absent and teardown-if-vanished. There was no liveness notion for existing records, and no classification of *permanent* vs *transient* start failures — everything took the retry/backoff path.

## Fix

1. **Liveness probing (#765).** Existing records of still-connected devices are probed behind a new `tunnelProber` seam, at most once per 30s per device (`defaultProbeInterval`). The default `rsdProber` is a 5s-timeout TCP dial of the tunnel's RSD endpoint using #808's `ios.DialTunnelTCPWithTimeout`; userspace tunnels are skipped (the localhost forwarder accepts connects regardless of device state, so a dial proves nothing). On probe failure the record is torn down via the per-device stop machinery from #738 and the create loop rebuilds it in the same cycle. Fresh tunnels defer their first probe by a full interval.
2. **Permanent-failure classification (#523).** `manualPairingTunnelStart` wraps its version-based failures (iOS <17, and userspace on 17.0–17.3) in a new `ErrUnsupportedVersion` sentinel. `UpdateTunnels` logs those **once at info** and marks the udid permanently skipped for the process lifetime (restart the agent after upgrading a device's OS). Transient errors keep the existing `failedDevice` exponential backoff untouched.
3. **udid filter (#607/#479).** The `--udid` filter path is now locked in by unit tests: only the matching device is attempted/tunneled, empty filter keeps managing all devices. No behavior change needed — the plumbing and `ios tunnel start` help text are already on main; this closes the issues with the behavior now guaranteed by tests.

Also adds a `productVersion` seam (defaults to `ios.GetProductVersion`) so the start path is unit-testable without a device.

## Options considered

- **Probe = short-timeout RSD TCP dial (chosen)** vs a full RSD/RemoteXPC handshake: the incident signature is a dead route (SYN blackhole), which a plain connect detects; a handshake adds protocol weight and failure modes for little extra signal. The `tunnelProber` interface leaves room to upgrade later.
- **Probe cadence:** every-cycle probing (1s) would dial each device every second and stall the loop up to 5s per dead device; a 30s per-device interval bounds overhead while detecting death well under a minute.
- **#523 blacklist on *any* first failure** (patch suggested in the issue thread): too aggressive — a transient pairing hiccup would permanently disable a healthy device. Classifying only version errors as permanent keeps backoff for everything genuinely retryable.
- **Pruning the unsupported set on disconnect** (device might be OS-upgraded): rejected for now — issues ask for silence; process-lifetime skip is the simplest contract and an agent restart re-detects.

## Test plan

New device-free tests in `ios/tunnel/tunnel_manager_robustness_test.go` (fake starter/lister/prober/version seams), each failing without its change:

- `TestUpdateTunnelsRebuildsDeadTunnel` — device still listed, prober says dead → old record closed exactly once, tunnel restarted, new record served.
- `TestUpdateTunnelsKeepsHealthyTunnel` — healthy record probed but untouched.
- `TestUpdateTunnelsDoesNotProbeDisconnectedDevice` — vanished devices stay with the disconnect teardown.
- `TestShouldProbeRespectsInterval`, `TestRsdProberSkipsUserspaceTunnels`.
- `TestManualPairingTunnelStartClassifiesUnsupportedVersions` — iOS 16.x and userspace-17.2 both classify as `ErrUnsupportedVersion`.
- `TestUpdateTunnelsSkipsUnsupportedDevicePermanently` — one attempt across three cycles, no backoff entry; `TestUpdateTunnelsKeepsBackoffForTransientFailures` — transient errors still back off, never marked unsupported.
- `TestUpdateTunnelsUdidFilter` — filter attempts only the matching device; empty filter manages all.

`go build ./...`, `go test ./...` and `go test -race ./ios/tunnel/` green; `gofmt -l` clean. Real-device e2e via `/test-devices`.

Fixes #765, fixes #523, fixes #607, fixes #479

### DIFF
--- a/ios/connect.go
+++ b/ios/connect.go
@@ -296,6 +296,42 @@ func initializeXpcConnection(h *http.HttpConnection) error {
 	return nil
 }
 
+// TunnelDialTimeout bounds TCP connects to tunnel/RSD endpoints. Without an
+// explicit timeout, a dial to a dead-but-still-routed tunnel address (device
+// rebooted or hung while the host-side TUN interface and route stayed up)
+// blocks for the kernel's TCP SYN timeout (~135s on Linux) per operation. 15s
+// is far above any healthy tunnel connect (sub-second) while still failing
+// fast enough for staleness handling to react.
+const TunnelDialTimeout = 15 * time.Second
+
+// ErrDialTimeout marks a tunnel/RSD TCP connect that exceeded go-ios' dial
+// timeout rather than failing outright. Callers can use errors.Is to treat the
+// endpoint as stale: the route existed but the device never answered, which is
+// the signature of a dead tunnel whose interface lingers.
+var ErrDialTimeout = errors.New("dial timed out")
+
+// DialTunnelTCP connects to a tunnel/RSD TCP endpoint (address in the form
+// accepted by net.Dial, e.g. "[fd00::1]:1234") with TunnelDialTimeout.
+func DialTunnelTCP(address string) (*net.TCPConn, error) {
+	return DialTunnelTCPWithTimeout(address, TunnelDialTimeout)
+}
+
+// DialTunnelTCPWithTimeout is DialTunnelTCP with a caller-chosen timeout.
+// Timeout errors are wrapped in ErrDialTimeout so they stay distinguishable
+// from refused/unreachable errors.
+func DialTunnelTCPWithTimeout(address string, timeout time.Duration) (*net.TCPConn, error) {
+	d := net.Dialer{Timeout: timeout}
+	conn, err := d.Dial("tcp", address)
+	if err != nil {
+		var netErr net.Error
+		if errors.As(err, &netErr) && netErr.Timeout() {
+			return nil, fmt.Errorf("%w after %v: %w", ErrDialTimeout, timeout, err)
+		}
+		return nil, err
+	}
+	return conn.(*net.TCPConn), nil
+}
+
 // ConnectTUNDevice creates a *net.TCPConn to the device at the given address and port.
 // If the device is a userspaceTUN device provided by go-ios agent, it will connect to this
 // automatically. Otherwise it will try a operating system level TUN device.
@@ -310,8 +346,7 @@ func ConnectTUNDevice(remoteIp string, port int, d DeviceEntry) (*net.TCPConn, e
 		return connectTUN(remoteIp, port)
 	}
 
-	addr, _ := net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%d", d.UserspaceTUNHost, d.UserspaceTUNPort))
-	conn, err := net.DialTCP("tcp", nil, addr)
+	conn, err := DialTunnelTCP(fmt.Sprintf("%s:%d", d.UserspaceTUNHost, d.UserspaceTUNPort))
 	if err != nil {
 		return nil, fmt.Errorf("ConnectUserSpaceTunnel: failed to dial: %w", err)
 	}
@@ -332,11 +367,7 @@ func ConnectTUNDevice(remoteIp string, port int, d DeviceEntry) (*net.TCPConn, e
 
 // connect to a operating system level TUN device
 func connectTUN(address string, port int) (*net.TCPConn, error) {
-	addr, err := net.ResolveTCPAddr("tcp6", fmt.Sprintf("[%s]:%d", address, port))
-	if err != nil {
-		return nil, fmt.Errorf("ConnectToHttp2WithAddr: failed to resolve address: %w", err)
-	}
-	conn, err := net.DialTCP("tcp", nil, addr)
+	conn, err := DialTunnelTCP(fmt.Sprintf("[%s]:%d", address, port))
 	if err != nil {
 		return nil, fmt.Errorf("ConnectToHttp2WithAddr: failed to dial: %w", err)
 	}
--- a/ios/connect_test.go
+++ b/ios/connect_test.go
@@ -1,7 +1,9 @@
 package ios
 
 import (
+	"errors"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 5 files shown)


danielpaulus:
Adversarial review + fixes (pushed to this branch as \`0352b30\`).

**Verdict:** logic was sound (probe lives outside the mux, map access is consistently mutex-guarded, \`%w\` classification and \`errors.Is\` work, teardown/\`lastProbe\` cleanup was complete, no orphaned utun). Fixed two confirmed defects and one bounded-growth issue in the self-heal path:

1. **Probe starvation (HIGH).** The liveness-probe loop ran serially inside \`UpdateTunnels\`; each dead-but-still-routed tunnel blackholes the SYN for the full \`probeDialTimeout\` (5s), so N dead tunnels could delay tunnel *creation* for newly connected devices by up to 5s×N. Probes now run **concurrently, bounded by \`maxConcurrentProbes\` (8)** via a semaphore + WaitGroup, collected under the mux.

2. **False-positive teardown (MEDIUM).** A single failed probe immediately tore down a healthy tunnel. Teardown now requires **\`probeFailureThreshold\` (2) consecutive failures**, reset on any success (new \`probeFailures\` map, mutex-guarded), so a momentary route hiccup/load spike can't destroy a healthy tunnel. Trade-off: a persistently dead tunnel is now rebuilt in ~2 probe intervals (~60s) instead of ~30s — documented on \`defaultProbeInterval\`.

3. **Bounded-growth / stale-after-upgrade (MEDIUM).** \`unsupportedDevices\` (and the new \`probeFailures\`) are now pruned for disconnected devices in the disconnect teardown, so they stay bounded by current device count, and a device that was unsupported → upgraded to a tunnel-capable iOS → reconnected is retried instead of skipped for the process lifetime.

New tests: two-failure threshold, transient-failure absorption (counter reset), unsupported-on-disconnect pruning + reconnect-retry. Made \`fakeProber\` thread-safe (probes are concurrent now).

**Dismissed:** the exact-\`17.4.0\` userspace-TUN boundary (\`GreaterThan(17.4.0)\`) is pre-existing and unchanged by this PR (only the error string was reworded). No transient-version misclassification risk: \`GetProductVersion\` either returns the real version or errors (→ transient-failure backoff, not the permanent set).

**Verification (both branches):** \`go build ./...\`, \`go test ./...\`, \`go test -race ./ios/tunnel/\` all green; \`gofmt -l\` clean.

*Note: #808 (base of this stack) was reviewed and confirmed clean — no changes needed there.*

PR #811, opened 2026-08-05
Add 'ios tunnel service install/uninstall/status' to run the tunnel agent as a managed OS service

### LINKED ISSUE #750: Add `ios tunnel service install/uninstall` to run the tunnel agent as a managed OS service (cloudflared-style)

Running the go-ios tunnel agent (`ios tunnel start`, especially the upcoming `--cloud` mode) persistently on a host today means hand-writing and maintaining an OS service. On Linux that's a systemd unit in `/etc/systemd/system`, placing the binary somewhere stable, capturing the right flags + env (`ORCHESTRATOR_URL`, `--userspace`, pair-record path, working dir), setting a restart policy, and `enable`-ing it at boot. This is manual, easy to get wrong (env not inherited, wrong working dir, missing restart), and has to be repeated on every host.

That cost scales badly for the cloud/device-farm use case, where the agent is a **fleet agent** running on many device hosts. Onboarding a host should be one command, not a unit-file checklist. (We hit exactly this friction bringing up a host by hand — the service didn't inherit the shell env, the working dir/pair-record path mattered, etc.)

A `service` group under the tunnel subcommand, modeled on **`cloudflared service install`** (and `ngrok service install`, `gitlab-runner install`):

```
ios tunnel service install   [--cloud] [--userspace] [--user] [other tunnel flags]
ios tunnel service uninstall
ios tunnel service status        # optional thin wrapper
```

- `install` registers go-ios as a managed OS service that runs `ios tunnel start …`, **capturing the current invocation's flags and relevant env** (`ORCHESTRATOR_URL`, pair-record path, etc.) into the service definition so it reproduces a known-good config, sets a restart policy, and enables it at boot. Idempotent (re-running overwrites + re-enables).
- `uninstall` stops, disables, and removes it.

Onboarding a cloud host then becomes:

```
ios login
ios tunnel service install --cloud --userspace
```

…and the host joins the fleet and stays joined across reboots.

- **Document a systemd unit (status quo).** Works, but manual and per-host; doesn't scale to a fleet and is the source of the friction above.
- **Hand-rolled, systemd-only installer.** Lowest dependency cost (we already know the exact unit), covers the Linux device hosts. Good MVP, but doesn't match the cross-platform precedents.
- **`github.com/kardianos/service`** (used by gitlab-runner, telegraf, etc.) abstracts systemd / launchd / Windows-SCM behind one API. Matches cloudflared/ngrok cross-platform from day one. Dependency lives in the `main` application package, **not** the `ios` library, so the library stays dep-clean.

Recommendation: take the cloudflared route via `kardianos/service` for cross-platform install/uninstall, or — if we want to stay minimal first — ship systemd-only and add launchd/Windows later.

**Additional context**

Precedents (the fleet-agent / tunnel-daemon category conventionally self-installs): **cloudflared** (`cloudflared service install`), **ngrok** (`ngrok service install`), **gitlab-runner** (`gitlab-runner install`). Server software (Caddy/Consul/Vault/Prometheus) instead documents a unit — go-ios cloud mode is the former category.

Things to decide during implementation:

1. **Privilege.** A *system* service needs root/admin. A `--userspace` cloud agent doesn't need root to *run*, so offer `--user` (systemd `--user` / LaunchAgent) for the no-root case — noting that user services need **lingering** enabled to run headless without a login session. Default to the system service (robust for headless farm hosts), document the sudo requirement.
2. **Env capture.** Bake `ORCHESTRATOR_URL` (and any other needed env) into the service definition (`Environment=` on systemd) — a service that doesn't inherit the operator's shell env silently runs with the wrong config.
3. **Consistency with `ENABLE_GO_IOS_AGENT`.** go-ios already self-spawns a tunnel child process (`RunAgent`, "experimental daemon mode"). The service command should complement that managed-process story, not contradict it.
4. **No-systemd Linux** (Alpine/OpenRC) — detect and fail with a clear message rather than writing a broken unit.
5. go-ios does **not** currently install any OS service; `selfIdentity.plist` is the only thing the binary writes itself.

Acceptance criteria (MVP):

- [ ] `ios tunnel service install [--cloud] [--userspace]` creates + enables + starts a service that survives reboot on Linux/systemd.
- [ ] The service reproduces the operator's flags and `ORCHESTRATOR_URL`.
- [ ] `ios tunnel service uninstall` fully removes it (stop, disable, delete).
- [ ] Clear errors when privileges are insufficient or no supported service manager is present.
- [ ] Docs note the sudo requirement and the `--user` alternative.

danielpaulus:
## Problem

Running the tunnel agent (`ios tunnel start`, and the upcoming `--cloud` fleet-agent mode) persistently on a host means hand-writing an OS service today: a systemd unit, the right flags and env (`ORCHESTRATOR_URL`, `--userspace`, pair-record path, working dir), a restart policy, and `enable`-ing it at boot. That is manual, easy to get wrong (env not inherited, wrong working dir, missing restart), and has to be repeated on every device host. Issue #750 asks for a cloudflared-style `service` group so onboarding a host is one command.

## Design

```
ios tunnel service install   [--userspace] [--user] [--pair-record-path=<p>] [--udid=<udid>] [--tunnel-info-host=<h>] [--tunnel-info-port=<p>]
ios tunnel service uninstall [--user]
ios tunnel service status    [--user]
```

- **install** captures the *current invocation* into a service definition that runs `ios tunnel start …`:
  - flags: `--pair-record-path` (resolved to an absolute path — relative paths are anchored to the install-time cwd, the `default` literal is passed through), `--userspace`, `--udid` (flag or `GO_IOS_UDID`), `--tunnel-info-host/port`;
  - env allowlist baked into the definition (`Environment=` on systemd / `EnvironmentVariables` on launchd): `ORCHESTRATOR_URL`, `GO_IOS_AGENT_HOST` — a service does not inherit the operator's shell env, so a known-good config must be captured explicitly;
  - `Restart=always`, enabled at boot, started immediately.
- **Idempotent**: re-running install stops/removes any existing definition and writes + re-enables the fresh one.
- **uninstall** stops, disables, and deletes the service; a service that isn't installed is reported, not an error. **status** prints installed/running state.
- **systemd-first**: on Linux, any other detected init system (OpenRC/upstart/SysV) fails with a clear message instead of writing an untested script. macOS launchd and the Windows service manager work through the same abstraction.
- **Privileges**: a system service needs sudo/admin; permission errors are decorated with actionable guidance (rerun with sudo, or pass `--user`). `--user` installs a systemd `--user` unit / launchd LaunchAgent for the no-root `--userspace` case; docs note the `loginctl enable-linger` requirement for headless hosts. Installing a user service without `--userspace` logs a warning (the kernel tunnel needs root).
- Dispatched as a **global command** (needs no device/usbmuxd), before the device commands so the `install`/`uninstall` literals don't collide with app (un)install.
- The `github.com/kardianos/service` dependency lives **only in package main**; the `ios` library packages stay dependency-clean.

## Implementation

- `cmd_tunnel_service.go` — config capture (`tunnelServiceConfig`), install/uninstall flows against a small `serviceController` interface (satisfied by `service.Service`, faked in tests), init-system gate, privilege-error decoration, command runner and output (JSON by default, `--nojson` human-readable).
- `cmd_global.go` — registers `tunnel service` in `globalCommands`.
- `main.go` — docopt usage line + detailed help text (sudo requirement, `--user` alternative, lingering note).
- `internal/clihelp/help.yaml`, `testdata/help/global.golden`, `README.md` — help catalog entry and docs.
- `go.mod`/`go.sum` — `github.com/kardianos/service v1.3.0` (root module only).

## Options considered

- **Document a systemd unit (status quo)** — works but stays manual and per-host; exactly the friction the issue describes.
- **Hand-rolled systemd-only installer** — lowest dependency cost and we know the exact unit, but Linux-only, and we'd own unit rendering, enable/disable, and privilege handling ourselves.
- **`github.com/kardianos/service`** *(preferred)* — one API over systemd/launchd/Windows SCM, battle-tested (gitlab-runner, telegraf), matches the cloudflared/ngrok precedent cross-platform from day one, and the dependency is confined to package main. We still gate Linux to systemd so we never ship an untested OpenRC/SysV script.

## Test plan

- Unit tests (device-free, no service is registered by tests):
  - `TestTunnelServiceConfigCapture` — golden files for flag+env capture: defaults, cloud/userspace/user, per-device agent, `GO_IOS_UDID` fallback, `default` pair-record literal (`testdata/tunnelservice/*.golden`).
  - Fake-injected failure paths: no systemd on Linux (`TestInstallTunnelServiceNoSystemd`), permission errors with sudo/`--user` hints (`TestInstallTunnelService{InsufficientPrivileges,UserServicePrivilegeError}`), idempotent reinstall ordering stop→uninstall→install→start (`TestInstallTunnelServiceIdempotent`), uninstall incl. not-installed (`TestUninstallTunnelService*`), dispatch matcher (`TestIsTunnelServiceCommand`).
- `go build ./...`, `go test ./...`, and `gofmt -l` clean; docopt parsing of the new usage line (incl. `--user` vs `--userspace`) verified.
- Manually verified read-only paths on macOS: `ios tunnel service status` (reports `not installed` via launchd) and `ios help tunnel service`.
- **Actual service registration (install → reboot survival → uninstall) is verified manually/e2e on the Linux device runner as a follow-up** — no service is installed by CI or unit tests.

Fixes #750


danielpaulus:
## Adversarial review (round 2)

Reviewed config-capture fidelity, install/uninstall state machine, kardianos platform quirks, and dispatch ordering. Verified behavior at runtime on macOS (launchd) and by inspecting the kardianos v1.3.0 templates. **No real services were left installed.**

### Confirmed defects — fixed in 45c8eef

1. **Windows service would never start (correctness bug).** The installed service runs `ios tunnel start` directly, but a Windows service must connect back to the service-control dispatcher (`svc.Run`) on startup — the tunnel agent never does, so the SCM kills it with error 1053 ("did not respond in a timely fashion"). `main.go` has no `service.Interactive()`/`svc.Run` wiring. The PR advertised Windows support and `checkServiceSystem` only gated non-systemd Linux, so `tunnel service install` on Windows would silently register a service that can never run. Fix: reject Windows in `checkServiceSystem` with an actionable message (pointing at nssm / scheduled task), matching the existing non-systemd-Linux gating; help text + README no longer promise Windows. (Proper SCM support is a larger, untestable-here change better left to a follow-up.)

2. **Incomplete env capture.** `GO_IOS_AGENT_HOST` was captured but not its partner `GO_IOS_AGENT_PORT` — both feed `ios.HttpApiHost/HttpApiPort`, so a custom agent port silently reverted to the default under the service (exactly the 'wrong config' failure the issue warns about). Also added `USBMUXD_SOCKET_ADDRESS` (ios/usbmuxconnection.go) so a non-standard usbmuxd socket is reproduced rather than falling back to default and seeing no devices. Golden + test updated.

### Verified correct (dismissed)

- **Dispatch ordering / install collision:** `tunnel service` is in `globalCommands` (dispatched before `deviceCommands`) and `isTunnelServiceCommand` requires both `tunnel` && `service`, so it can't collide with app `install`/`uninstall`. Confirmed by parsing + running the built binary.
- **Idempotent reinstall (stop→uninstall→install→start):** correct for same-scope ops; matches unit test. Ignoring `Stop` errors before uninstall is deliberate and standard.
- **launchd restart policy:** `RunAtLoad=true` + kardianos `KeepAlive` default true gives start-at-load + restart-on-crash. Correct.
- **systemd `Restart=always` + user unit:** kardianos hardcodes `WantedBy=multi-user.target`, which is valid in the user manager too; lingering caveat in the help is accurate.
- **`--enabletun`** appears in help text but is not read by `tunnel start` (dead doc flag) — correctly not captured.
- **kardianos dependency** stays in package main only.
- Note: mixing scopes (installing a *system* service while a *user* service is loaded) can confuse launchd's non-scoped `launchctl list`, but that's an inherent kardianos/launchctl limitation, not a PR-fixable state-machine bug.

### Acceptance criteria
- [x] install/enable/start survives reboot on Linux/systemd
- [x] reproduces flags + ORCHESTRATOR_URL (now also agent port + usbmuxd socket)
- [x] uninstall fully removes (stop/disable/delete); not-installed is a no-op
- [x] clear errors on insufficient privileges / unsupported manager (now incl. Windows)
- [x] docs note sudo + `--user` + lingering

PR #823, opened 2026-08-11
fix(accessibility): consume hostInspectorMonitoredEventTypeChanged continuously to prevent DTX deadlock

sakhisheikh:
## Problem

When driving the AX inspector (`ios/accessibility`), all element-change callbacks can permanently stop while outgoing commands (`Move`) keep working. The connection is alive, nothing errors — the client just never receives another event.

**Root cause:** the device emits `hostInspectorMonitoredEventTypeChanged:` not only as a response to `deviceInspectorSetMonitoredEventType`, but also when the device is physically touched — and on iOS 27, when tapping via XCUITest/WebDriverAgent. The selector was registered but only read one-shot inside `EnableSelectionMode`/`TurnOff`. An event arriving outside those windows blocked forever on the unbuffered channel send in `Channel.Dispatch`, which runs on the connection's single reader goroutine — halting delivery for every channel on that connection.

**Reproduction:** enable the inspector, navigate (works), physically tap the device (or on iOS 27, tap via WDA), navigate again → cursor still moves but no `hostInspectorCurrentElementChanged:` callbacks ever arrive.

## Fix

- Add a dedicated consumer goroutine for `hostInspectorMonitoredEventTypeChanged:`, wired up in `init` exactly like the existing `hostAppStateChanged:` / `hostInspectorNotificationReceived:` readers.
- Remove the one-shot awaits in `EnableSelectionMode`/`TurnOff` — with the consumer goroutine present they would race it for the same channel and hang. They only logged the value.
- Add `DisableSelectionMode()`: starting with iOS 27, while selection mode (monitored event type 2) is armed the AX daemon consumes touches to move inspector focus and the app never receives the touch — synthesized taps only highlight the element. Callers need to disable selection mode before synthesizing taps.

Verified on-device against iOS 26 and iOS 27: callbacks survive physical touches and WDA taps, and taps activate elements on the first try after disabling selection mode.

Related: #824 hardens `Channel.Dispatch` against this whole class of bug (never block the reader goroutine); this PR fixes the specific missing consumer and is sufficient on its own.


danielpaulus:
Thanks @sakhisheikh — nicely written up, and the root-cause analysis holds up under review.

## What this does
Adds a dedicated `readhostInspectorMonitoredEventTypeChanged` goroutine that continuously drains `hostInspectorMonitoredEventTypeChanged:`, wired into `init` exactly like the existing `readhostAppStateChanged` / `readhostInspectorNotificationReceived` readers; removes the one-shot `awaitHostInspectorMonitoredEventTypeChanged` calls from `EnableSelectionMode`/`TurnOff` (which only logged); and adds `DisableSelectionMode()`.

## Correctness — confirmed
The deadlock is real. `Channel.Dispatch` does an unbuffered `v <- msg` (`ios/dtx_codec/channel.go:179`) for any registered selector, and that runs inline on the connection's single reader goroutine (`ios/dtx_codec/connection.go:200-220`). Registering `hostInspectorMonitoredEventTypeChanged:` (`accessibility_control.go:122`) without a continuous consumer means any event arriving outside the old one-shot window blocks the reader and stalls *all* delivery on the connection. The fix is the idiomatic one for this file and strictly improves on one-shot consumption. Dropping the awaits is correct — with the drain goroutine present a second reader on the same channel would race it. No selector-stealing risk: each selector has its own channel, so the drain can't consume `hostInspectorCurrentElementChanged:`.

## Build / test
`go build ./...`, `go vet ./ios/accessibility/...`, and `go test ./ios/accessibility/...` all pass on the branch. No unit tests for the new goroutine, which is fine — this is device-observable behavior and belongs in the `e2e` suite per AGENTS.md, not package unit tests. The on-device verification against iOS 26/27 in the description covers it.

## One optional polish (non-blocking)
The new goroutine drops the unarchive error and then indexes the result:
```go
n, _ := nskeyedarchiver.Unarchive(msg.Auxiliary.GetArguments()[0].([]byte))
golog.Info(..., "value", n[0])
```
Its two sibling readers (`readhostAppStateChanged`, `readhostInspectorNotificationReceived`) check that error before indexing. Since this is a purely informational log line, consider matching them — on a decode failure `n` is nil and `n[0]` panics the process, which also leaves the channel without a consumer (re-triggering the very deadlock this PR fixes). A guarded `if err != nil { continue }` (or dropping the `n[0]` value entirely) keeps it consistent and can't take the connection down.

## Notes for maintainer context (not asks on this PR)
- The broader \"ctx cancelled but connection still open ⇒ next event blocks the reader\" edge is a pre-existing property shared by the existing readers, and #824 hardening `Channel.Dispatch` is the right home for it. In the current CLI path (`runAxAudit` uses a `WithTimeout` ctx with `defer conn.Close()`) the connection tears down together with the ctx, so it isn't hit here.
- `DisableSelectionMode` (like the surrounding async setters) fires `deviceInspectorSetMonitoredEventType(0)` without awaiting an ack, so callers synthesizing a tap immediately after have no completion barrier. That matches the existing style in this file and is out of scope, but worth keeping in mind if tap-after-disable turns out flaky in practice.

## Recommendation
Looks good — mergeable as-is. The unarchive-error guard is the only thing I'd suggest tightening, and it's optional. Not approving here since real-device e2e (`/test-devices`) is the meaningful gate for accessibility changes.

_— automated review pass (Claude), on the maintainer's behalf; a human should confirm before merge._

danielpaulus:
🎉 Merging this — thank you so much, @sakhisheikh! 🙏

This is a genuinely great catch and fix. The accessibility inspector deadlock was subtle: registering `hostInspectorMonitoredEventTypeChanged` without a continuous consumer blocks the dtx connection's single reader goroutine, which silently wedges *every* channel on that connection ("commands send, no callbacks arrive"). Your continuous-drain goroutine is exactly the right, idiomatic fix, and adding `DisableSelectionMode()` rounds it out nicely. We reviewed it end-to-end against the dtx codec internals and it holds up — clean, well-scoped, and on-device verified against iOS 26/27.

Really appreciate you taking the time to track this down and contribute it back. Contributions like this make go-ios better for everyone. Welcome, and thank you! 🚀

(One tiny optional follow-up if you're ever in the area: the new drain goroutine indexes `n[0]` without checking the `Unarchive` error first — a one-line `if err != nil { continue }` guard would harden it against a malformed decode. Totally non-blocking; merging as-is.)

danielpaulus (approved):
Reviewed against the dtx codec internals — correct fix for the reader-goroutine deadlock. Thanks again, @sakhisheikh! Merging.

PR #824, opened 2026-08-11
fix(dtx): never block the connection reader goroutine in Channel.Dispatch

sakhisheikh:
## Problem

`Channel.Dispatch` runs on the DTX connection's single reader goroutine. Three code paths could block it forever, halting delivery for **every** channel on the connection — head-of-line blocking that presents as a silent hang (outgoing commands keep working, incoming messages are never delivered again):

1. **Registered-selector events** were sent on unbuffered channels (`v <- msg`). An event arriving while no `ReceiveMethodCall` was pending blocked forever. This is how the AX inspector deadlock in #823 manifested.
2. **Late replies**: `sendAndAwaitReply` never deregistered its waiter on context timeout. A reply arriving after the caller gave up was sent into the abandoned unbuffered channel — blocking forever.
3. **Unknown replies**: a reply with no registered waiter was sent on a nil map entry, which blocks forever.

## Fix

Delivery is now always guarded — the reader goroutine can never block:

- Per-selector queues are bounded (16); on overflow the **oldest** event is dropped with a warning so the most recent state (e.g. current AX cursor position) survives for a consumer that returns.
- Response waiters are buffered(1) and deregistered on timeout/send failure; late or unknown replies are dropped with a warning.
- The fragment path uses the same guarded delivery.

A note on "buffering just delays the problem": agreed — the buffer here only absorbs bursts and is not the fix. The load-bearing change is the guarded, non-blocking send: absence of a consumer is a protocol-usage bug, and it is now **observable** (a log line naming the selector/identifier) instead of a connection-wide silent hang.

## Tests

`channel_robust_test.go` locks in the guarantees:
- Dispatch does not block with zero consumers, even at 4× buffer volume
- overflow drops oldest, newest survives
- a late reply after caller timeout does not block
- a reply with an unknown identifier does not block

Found while debugging the AX inspector callback loss in #823 — that PR fixes the specific missing consumer; this one makes the dispatcher robust against the whole class of bug.

### DIFF
--- a/ios/dtx_codec/channel.go
+++ b/ios/dtx_codec/channel.go
@@ -36,10 +36,20 @@ func WithTimeout(seconds uint32) ChannelOption {
 	}
 }
 
+// registeredMethodBuffer bounds how many not-yet-consumed remote method calls
+// are kept per selector. Dispatch runs on the connection's single reader
+// goroutine and must never block: without a buffer, one event arriving while
+// no ReceiveMethodCall is pending would stall delivery for every channel on
+// the connection (head-of-line blocking with a silent hang).
+const registeredMethodBuffer = 16
+
 func (d *Channel) RegisterMethodForRemote(selector string) {
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
-	d.registeredMethods[selector] = make(chan Message)
+	if d.registeredMethods == nil {
+		d.registeredMethods = map[string]chan Message{}
+	}
+	d.registeredMethods[selector] = make(chan Message, registeredMethodBuffer)
 }
 
 func (d *Channel) ReceiveMethodCall(selector string) Message {
@@ -156,18 +166,52 @@ func (d *Channel) sendAndAwaitReply(ctx context.Context, expectsReply bool, mess
 	if err != nil {
 		return Message{}, err
 	}
-	responseChannel := make(chan Message)
+	// Buffered so a reply racing with our timeout can always be deposited by
+	// sendResponse instead of being dropped on the non-blocking send.
+	responseChannel := make(chan Message, 1)
 	d.AddResponseWaiter(identifier, responseChannel)
 
 	err = d.connection.Send(bytes)
 	if err != nil {
+		d.removeResponseWaiter(identifier)
 		return Message{}, err
 	}
 	select {
 	case response := <-responseChannel:
 		return response, nil
 	case <-ctx.Done():
+		// Deregister so the abandoned waiter does not leak in responseWaiters
+		// (and a stale defragmenter for this identifier is cleaned up too).
+		d.removeResponseWaiter(identifier)
 		return Message{}, fmt.Errorf("Timed out waiting for response for message:%d channel:%d", identifier, d.channelCode)
+	}
+}
+
+func (d *Channel) removeResponseWaiter(identifier int) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	delete(d.responseWaiters, identifier)
+	delete(d.defragmenters, identifier)
+}
+
+// deliverEvent hands a device-initiated method call to the selector's queue
+// without ever blocking the reader goroutine. The queue is LOSSY BY DESIGN:
+// when it is full the OLDEST event is dropped (with a warning) so the most
+// recent state survives. This is the right trade-off for the current
+// latest-state consumers (e.g. the AX inspector's cursor position); do not
+// route a selector through here if every single callback matters.
+func (d *Channel) deliverEvent(queue chan Message, msg Message, selector string) {
+	for {
+		select {
+		case queue <- msg:
+			return
+		default:
+		}
+		select {
+		case dropped := <-queue:
+			golog.Warn("event queue full, dropping oldest event", "module", logModule, "channel_id", d.channelName, "selector", selector, "dropped_identifier", dropped.Identifier)
+		default:
+		}
 	}
 }
 
@@ -184,7 +228,10 @@ func (d *Channel) Dispatch(msg Message) {
 		golog.Trace("dispatching", "module", logModule, "channel_id", d.channelName, "selector", selector)
 		if v, ok := d.registeredMethods[selector]; ok {
 			d.mutex.Unlock()
-			v <- msg
+			// Never block the reader goroutine: an event arriving while no
+			// ReceiveMethodCall is pending would otherwise stall delivery for
+			// every channel on this connection.
+			d.deliverEvent(v, msg, selector)
 			return
 		}
 	}
--- a/ios/dtx_codec/channel_robust_test.go
+++ b/ios/dtx_codec/channel_robust_test.go
@@ -1,0 +1,111 @@
+package dtx
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+// newTestChannel is shared with hardening_channel_internal_test.go.
+
+func methodInvocationMessage(identifier int, selector string) Message {
+	return Message{
+		Identifier:    identifier,
+		PayloadHeader: PayloadHeader{MessageType: Methodinvocation},
+		Payload:       []interface{}{selector},
+	}
+}
+
+// Dispatch must never block the connection's reader goroutine, even when a
+// registered selector has no active ReceiveMethodCall consumer.
+func TestDispatchDoesNotBlockWithoutConsumer(t *testing.T) {
+	channel := newTestChannel()
+	channel.RegisterMethodForRemote("someEvent:")
+
+	done := make(chan struct{})
+	go func() {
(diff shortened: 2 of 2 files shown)


danielpaulus:
Automated adversarial review pass (dtx concurrency). Not an approval — flagging for a human maintainer. Thanks for the careful writeup, @sakhisheikh; the diagnosis is genuinely useful and the core insight is correct.

## What the change does
`Channel.Dispatch` runs on the connection's single reader goroutine (`reader()` in `connection.go`, one `go reader(...)` per connection), so any blocking send there stalls delivery for every channel on that connection. The PR makes three delivery paths non-blocking:
- registered-selector events → bounded (16) per-selector queue with drop-oldest (`deliverEvent`)
- response waiters → buffered(1) + deregistered on timeout/send-failure, non-blocking send (`deliverResponse`)
- fragment reassembly → routed through the same guarded delivery, and the mutex is no longer held across `Extract`/decode/dispatch

The premise is real and the registered-selector fix is the load-bearing one.

## Build / test / race (on the PR branch)
- `go build ./...` ✅
- `go test ./ios/dtx_codec/...` ✅
- `go test -race ./ios/dtx_codec/...` ✅ (added tests are good: no-consumer non-block, drop-oldest, late reply, unknown reply)

## Blocking concern — this needs a rebase, and the rebase is not mechanical

The PR was branched from an older `main` and **conflicts**. Crucially, `main` already landed the reply-side non-blocking fix in the input-hardening PR (#791): the `sendResponse(waiter, msg)` helper (nil-guard + non-blocking `select`/`default`) plus a guarded selector extraction. So parts 2 and 3 of this PR ("late replies", "unknown replies") are **already fixed on main** by `deliverResponse`'s equivalent — those need to be reconciled with `sendResponse`, not re-implemented.

More seriously, this PR **regresses a security fix from #791**:

- `channel.go:241-244` — `msg.Payload[0].(string)` is used unguarded (twice). `main` deliberately replaced this with `len(msg.Payload) > 0` + comma-ok assertion, precisely because a `Methodinvocation` frame from an untrusted device can carry an empty or non-string payload. I confirmed the panic locally: a `Methodinvocation` with an empty payload panics `index out of range [0] with length 0`. Worse, the panic fires at line 241 **while `d.mutex` is still held** (locked at 236, no unlock yet) inside the un-recovered reader goroutine → it crashes the process and would leave the channel mutex locked. This is a real DoS regression of #791 and the single most important thing to fix before this can go in.

## Correctness of the non-blocking design (once rebased)

The drop-oldest event queue is sound: Dispatch is the *only* producer (single reader goroutine), so the drain-then-insert loop is FIFO and the "newest survives" guarantee holds (your test proves it). Worth being explicit that this is a **lossy** tradeoff: for the AX inspector's "latest cursor state" selectors, dropping the oldest is the right call; but any selector where every callback matters (not just latest state) would silently lose events past 16 with only a WARN. The class of consumers today (`accessibility_control.go`) all look latest-state, so this is acceptable — please just call the lossiness out in the queue comment so a future selector author knows the queue is not lossless.

## Lower-priority notes (pre-existing, not blockers)
- `removeResponseWaiter` (timeout/send-fail path) deletes only `responseWaiters`, not `defragmenters`; a reply that started fragmenting and never completes leaks its `defragmenters[id]` entry. This mirrors main's behavior, so not introduced here, but while you're in this code it'd be nice to also drop the defragmenter on abandon.
- Fragment path continues to use `fullMsg` after `DecodeNonBlocking` returns `err`/leftover (only logs). Same as main — pre-existing, worth a follow-up but out of scope here.
- Codex flagged a `FragmentDecoder` data race from concurrent `Dispatch`; I checked and it's **not reachable** — there is exactly one reader goroutine per connection, so per-channel Dispatch is serialized. No race (confirmed by `-race`).

## Recommendation
**Needs changes before merge.** The non-blocking direction and the registered-selector fix are correct and welcome, and `-race` is clean. But: (1) rebase onto current `main` and reconcile with the existing `sendResponse` reply-path fix rather than duplicating it, and (2) restore the guarded selector extraction from #791 — the current unguarded `msg.Payload[0].(string)` reintroduces a device-triggerable reader-goroutine panic (with the mutex held) that #791 specifically closed. Once those are addressed this is close.

— automated review pass; a human maintainer should confirm before merge/approve.

danielpaulus (changes requested):
Thanks for this, @sakhisheikh — and thank you again for #823! The core idea here is right: `reader()` is a single goroutine per connection that calls `Channel.Dispatch` synchronously, so a blocking send genuinely can deadlock the whole connection, and your bounded-queue + drop-oldest `deliverEvent` is a sound, well-tested fix (nice `-race` coverage). I'd love to get this in — two things need to change first:

**1. Blocker — please restore the guarded payload extraction.** `channel.go` uses an unguarded `msg.Payload[0].(string)`, which reverts the input-hardening from #791. An empty-payload message from an untrusted device panics with `index out of range [0]` — and it fires while `d.mutex` is held in the un-recovered reader goroutine, so it crashes the process and leaks the lock (a device-triggerable DoS). Please keep #791's `len(msg.Payload) > 0` + comma-ok extraction.

**2. Rebase onto current main + reconcile the reply path.** This branched before #791, so it conflicts, and its `deliverResponse` duplicates the `sendResponse` (nil-guard + non-blocking select/default) that already landed on main. After rebasing, the 'late replies' / 'unknown replies' parts are largely already solved — please fold your changes into the existing `sendResponse` rather than adding a parallel path.

Once those are addressed the concurrency fix itself looks good to me. One small note for future selector authors (fine as-is): the registered-event queue is lossy by design (drop-oldest with a WARN), which is correct for the current latest-state AX consumers. Happy to re-review as soon as you push the rebase. Thanks again! 🙏

_(automated review pass on the maintainer's behalf — a human will confirm before merge)_

sakhisheikh:
Thanks for the thorough review — both points addressed, rebased onto current `main` (9c9fa2a):

1. **Guarded payload extraction restored**: the branch is now based on main, so #791's `len(msg.Payload) > 0` + comma-ok extraction is intact (the previous revert was an artifact of branching from v1.2.1 — apologies).
2. **Reply path folded into `sendResponse`**: no more parallel `deliverResponse`. What remains on top of main's reply handling: the waiter is now buffered(1) and **deregistered on timeout/send-failure** — abandoned waiters no longer leak in `responseWaiters`, a reply racing the timeout can be deposited instead of dropped, and per your note the deregistration also drops the stale `defragmenters` entry for the abandoned identifier.
3. The queue comment now calls out the **lossy-by-design** drop-oldest semantics explicitly for future selector authors.

The diff is now just the load-bearing piece: non-blocking `deliverEvent` for registered selectors + waiter lifecycle. `go build`, `go vet`, and `go test -race ./ios/dtx_codec/` all pass (the new tests reuse `newTestChannel` from the #791 hardening tests).

danielpaulus (approved):
Thank you for this, @sakhisheikh — it's a real, still-needed fix and it's more valuable than it first looks. 🙏

We ran it through a careful cross-review, and here's what it actually buys us on top of what's already on `main`:

- **It fixes the #823 deadlock that's still live.** `Channel.Dispatch` still does a blocking `v <- msg` onto an *unbuffered* `registeredMethods` channel on the single connection reader goroutine — so one event arriving with no `ReceiveMethodCall` pending stalls delivery for every channel on the connection. Your buffered queue + non-blocking `deliverEvent` (drop-oldest) is the right fix; the earlier hardening PRs (#792/#832) hardened the *response* path and the *global* dispatcher, but never this per-channel registered-method send.
- **It also closes a reply-drop race we'd left behind.** #792 made `sendResponse` non-blocking, which — onto an unbuffered channel — can silently drop a fast reply that lands before the caller parks in its `select`, turning a good call into a spurious timeout. Your buffered(1) `responseChannel` closes that window.
- **And it stops a leak.** `removeResponseWaiter` on timeout / send-error cleans up the `responseWaiters` + `defragmenters` entries that were otherwise leaking on every timed-out call.

We double-checked the concurrency (the drop-oldest loop, lock ordering, the buffered-timeout race) — no issues; `Dispatch` is single-producer so the lossy queue is safe, and all `RegisterMethodForRemote` consumers are latest-state accessibility events, which is exactly the right use for drop-oldest.

The only red check is the unrelated iOS 26 WebInspector/accessibility flake on the macOS runner — nothing to do with this change (unit tests are green). Merging now.

One small follow-up we'll handle ourselves in a separate PR (no action needed from you): trimming a duplicate test (`TestUnknownReplyDoesNotBlock` overlaps existing coverage) and reshaping `TestLateReplyAfterTimeoutDoesNotBlock` to assert the waiter/defragmenter cleanup. Keeping your two queue tests as-is — they're the valuable new coverage.

Great contribution — thanks again. 🚀

PR #831, opened 2026-08-12
test(e2e): retry preios17 device commands on transient usbmuxd/transport drops

danielpaulus:
## Problem

On the `ganjalf` Linux runner the pre-iOS17 e2e device (udid `5b98c87e742570f10e22113b7dd8502638011e16`, iOS 12.5.7) intermittently **drops off usbmuxd mid-suite**. When it does, five tests in `test/e2e/preios17` fail hard with no retry, red-walling the entire PR fleet on a single infra hiccup:

- `TestAccessibilityAudit`
- `TestPs`
- `TestScreenshot`
- `TestLaunchKill`
- `TestSyslog`

The failures are always **transient transport-layer errors**, never assertion or genuine command failures, e.g.:

- `write unix @->/var/run/usbmuxd: write: broken pipe`
- `DeviceConnection failed writing all bytes ... sent: 0`
- `could not connect to RSD ... got RST frame with error code: STREAM_CLOSED`
- dtx `Timed out waiting for response for message:5 channel:0`

## Fix

This mirrors the **already-accepted pattern** in the tunnel screenshots strand (`test/e2e/tunnel/perf_test.go`), which retries the same class of transient RST/`STREAM_CLOSED` connection resets up to 4×.

A bounded retry is added to the harness (`RunForDeviceResilient` / `SmokeArrResilient` / `AuditAfterLaunchResilient` / `StreamSmokeResilient`). It:

- runs the command through the existing **bounded** runner (`runBounded`), so a wedge is still bounded and stack-dumped;
- retries **only** when stderr matches a known transient-transport signature (case-insensitive), among: `broken pipe`, `sent: 0` / `sent:0`, `could not connect to RSD`, `failed to get tunnel info`, `STREAM_CLOSED`, `connection reset`, `Timed out waiting for response`;
- retries at most **3 attempts** with **~2s backoff**, logging each retry with the matched signature via `t.Logf`;
- **fails immediately** on any other error (or empty output without a transient signature) — real bugs are never masked;
- after exhausting retries, `t.Fatalf`s with the last error.

The signature list lives in **one place** in `test/e2e/harness/harness.go` with a comment scoping it strictly to the flaky pre-iOS17 device's transport drops. **No assertion is weakened** — the schema/value checks (non-empty JSON array + element keys, PNG magic + size, audit issueType, streamed-output-within-window) are byte-for-byte the same.

This is **test-only** code under `test/e2e/` (build tag `e2e`); no production package is touched.

## Options considered

- **Bounded retry here (chosen).** Cheap, self-healing, scoped to a known-transient signature set, mirrors an existing accepted strand, and never masks a real failure. Downside: adds up to ~6s per affected command in the rare drop case.
- **Fix the physical device / cable.** The real root cause is a flaky USB link on that specific device. Worth doing, but it's a hardware/ops task on the self-hosted runner and can't gate PRs in the meantime; this change is orthogonal and still desirable once the cable is replaced.
- **Quarantine the device.** Dropping the pre-iOS17 device from CI would lose the *only* real-device coverage of the direct-over-usbmuxd DTX/instruments path (ps/screenshot/launch/kill without a tunnel), which is exactly what this suite exists to prove. Rejected.

## Verification

`go build -tags e2e ./test/e2e/...`, `go vet -tags e2e ./test/e2e/...`, and `gofmt -l` are all clean. The real-device workflow was dispatched on this branch to confirm the five preios17 tests still PASS on the healthy device (no regression); farm result linked in the PR thread.

### DIFF
--- a/test/e2e/harness/harness.go
+++ b/test/e2e/harness/harness.go
@@ -178,6 +178,184 @@ func RunForDevice(t *testing.T, udid string, args ...string) []byte {
 	return RunIOS(t, append(args, "--udid="+udid)...)
 }
 
+// transientTransportSignatures are the stderr fragments (matched
+// case-insensitively) that identify a *transient* usbmuxd / USB-transport drop
+// on the flaky pre-iOS17 CI device (udid 5b98c87e…, iOS 12.5.7). That device
+// intermittently falls off usbmuxd mid-suite; when it does, an otherwise-good
+// command fails with one of these transport-layer errors — never with an
+// assertion or a genuine command error. Retrying ONLY on these signatures lets a
+// single infra hiccup self-heal without masking real bugs. Keep this list here,
+// in one place; do not broaden it to cover anything but known transport drops.
+var transientTransportSignatures = []string{
+	"broken pipe",                    // write unix @->/var/run/usbmuxd: write: broken pipe
+	"sent: 0",                        // DeviceConnection failed writing all bytes ... sent: 0
+	"sent:0",                         // same, no space
+	"could not connect to rsd",       // got RST frame with error code: STREAM_CLOSED
+	"failed to get tunnel info",      // tunnel/RSD info lost when the device drops
+	"stream_closed",                  // HTTP/2 RST frame after the connection resets
+	"connection reset",               // generic peer/USB reset
+	"timed out waiting for response", // dtx: Timed out waiting for response for message:N channel:0
+}
+
+// matchTransientTransport reports the first transient-transport signature found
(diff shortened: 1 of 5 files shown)


danielpaulus:
Closing per maintainer preference: we'll just re-run the real-device CI when the pre-iOS17 device transiently drops off usbmuxd, rather than add transient-transport retry logic to the preios17 tests. Keeps the e2e tests simple. (The underlying flake is a physical device/USB drop on the Linux runner, not a code issue.)

PR #832, opened 2026-08-12
kill: add --watch to kill blocked apps the instant they launch

aluedeke:
## What

Adds `--watch` to `ios kill` — stays resident and kills each given bundleID/`--process` the instant it's next launched, blocking it from running. Also lets `ios kill` take multiple bundleIDs at once (was single-bundleID only).

```
ios kill (<bundleIDs>... | --pid=<processID> | --process=<processName>) [--watch] [options]
```

## Bug fixes required to make this work

`--watch` builds on `ListenAppStateNotifications`, which already shipped behind `ios instruments notifications` but was non-functional (blocked forever). Fixed:

- `GlobalDispatcher.Dispatch` (`ios/dtx_codec/connection.go`) didn't forward unsolicited `Methodinvocation`-type pushes on the global channel to a connection's `MessageDispatcher`.
- `ListenAppStateNotifications` (`ios/instruments/instruments_notifications.go`) listened on the wrong channel and used an unbuffered message channel, deadlocking its own setup calls.
- `needsAutomaticTunnelInfo` (`cli_device_resolution.go`) didn't gate RSD tunnel-info enrichment for the new watch path.
- `matchesKillTarget` (`cmd_device_apps.go`): multi-bundleID kill used `pid == 0` as "no `--pid` given", which also matched the Mach kernel's real pid-0 entry.

`WatchKill` (`ios/instruments/watchkill.go`) shares one DTX connection for both the notification and kill channels, rather than opening two.

## Validation

Live against a physical iPhone SE (iOS 26.6): multi-bundleID kill, `--watch` killing a target within single-digit milliseconds of launch while an unguarded app is left running, and the pid-0 fix confirmed not to affect device stability. `go test ./...` clean aside from two pre-existing, unrelated `ios/imagemounter` failures.


danielpaulus (approved):
Thank you so much for this, @qawolf — this is a genuinely excellent contribution. 🙏

`ios kill --watch` is a great feature on its own, but what makes this PR stand out is that you didn't just bolt it on — you went and fixed the real plumbing underneath:

- **Push-based, not polling.** Subscribing to `applicationStateNotification:` means the launch→kill window is wire latency, not a poll interval — no busy-loop, no wasted CPU. Exactly right.
- **You closed a long-standing `dtx_codec` gap.** That `// TODO: use the dispatchFunctions map` was load-bearing: handling known selectors and forwarding unsolicited `Methodinvocation` pushes to the connection dispatcher is precisely what was silently dropping device pushes. The comments explaining *why* (deadlock avoidance on the unbuffered channel, avoiding dispatcher spam) are exactly what a future reader needs.
- **The concurrency is careful.** The single reader keeps draining pushes into the 256-buffered channel while `kill()` blocks on its own response channel with a timeout — the unbuffered version deadlocked, and you got the buffering + separation right.
- **You caught the sharp edges.** The pid-0 / Mach-kernel guard and the docopt `[]string` handling are both the kind of thing that bites later — and you handled *and* unit-tested them (splitting `matchesKillTarget`/`shouldWatchKill` out to be device-free testable is exactly the right call for this codebase).

Clean SIGINT shutdown, correct logging conventions, no library-package `slog`/`logrus` leakage. Polished work.

One tiny, non-blocking nit for a future pass: `runKillWatch` discards the `bundleID` from `shouldWatchKill`, so the kill log shows `process` but not the friendlier bundle id. Purely cosmetic.

The only red check (macOS `TestWebInspectorBrowserControl`) is a known WebInspector/CDP flake on our iOS 26 device runner — unrelated to this change; the Linux e2e suite passed in full — so I'm merging over it.

Approving and merging. Thanks again — this is a great one. 🚀

PR #837, opened 2026-08-13
Allow for Silent Install of Wifi Profile

outcast36:
The `ios wifi` command would previously only install the generated profile as if the device were unsupervised. Additionally, even if this was used with a supervised device, the check for supervision would always fail. 

Updates wifi command to take a `--p12file` and a `--p12password` similar to the `ios profile add` command. This allows for supervised devices to silently accept the profile.

Proposed wifi command structure:
```ios wifi --ssid=my_ssid --password=my_password --enc-type=my_encryption_type --p12file=certificate.p12 --p12password=a```

Tested with an iPhone 16 running iOS 26.2.1 on Fedora 39. Supervision was setup using ```ios prepare```

### DIFF
--- a/cmd_device_management.go
+++ b/cmd_device_management.go
@@ -50,6 +50,8 @@ func runWifiCommand(ctx commandContext) {
 	psw, _ := ctx.Args.String("--password")
 	encType, _ := ctx.Args.String("--enc-type")
 	remove, _ := ctx.Args.Bool("--remove")
+	p12file, _ := ctx.Args.String("--p12file")
+	p12password, _ := ctx.Args.String("--p12password")
 
 	if encType == "" {
 		encType = "WPA"
@@ -58,7 +60,8 @@ func runWifiCommand(ctx commandContext) {
 	if remove {
 		exitIfError("failed removing wifi", mcinstall.RemoveWifi(ctx.Device, ssid))
 	} else {
-		exitIfError("failed preparing wifi", mcinstall.PrepareWifi(ctx.Device, ssid, psw, encType))
+		exitIfError("failed preparing wifi", mcinstall.PrepareWifi(ctx.Device, ssid, psw, encType,
+			p12file, p12password))
 	}
 	fmt.Print(convertToJSONString("ok"))
 }
--- a/ios/mcinstall/wifi.go
+++ b/ios/mcinstall/wifi.go
@@ -126,7 +126,11 @@ func isSupervised(conn *Connection) bool {
 	if err != nil {
 		return false
 	}
-	s, _ := cfg["IsSupervised"].(bool)
+	cloudCfg, ok := cfg["CloudConfiguration"].(map[string]interface{})
+	if !ok {
+		return false
+	}
+	s, _ := cloudCfg["IsSupervised"].(bool)
 	return s
 }
 
--- a/ios/mcinstall/wifi.go
+++ b/ios/mcinstall/wifi.go
@@ -39,7 +39,7 @@ func PrepareWifi(device ios.DeviceEntry, ssid string, psw string, encType string
 	golog.Debug("flush response", "module", logModule, "udid", device.Properties.SerialNumber, "response", re)
 
 	supervised := isSupervised(conn)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by outcast36** [comment]
From what I can tell, the failing tests are unrelated to changes in these commits. Those being:
```--- FAIL: TestAccessibilityAudit (0.00s)
    --- PASS: TestAccessibilityAudit/00008120-001905DE2E9B401E (2.25s)
    --- FAIL: TestAccessibilityAudit/00008110-001C58C00A88401E (60.96s)
```
and
```
--- FAIL: TestWebInspectorBrowserControl (0.00s)
    --- SKIP: TestWebInspectorBrowserControl/00008120-001905DE2E9B401E (0.15s)
    --- FAIL: TestWebInspectorBrowserControl/00008110-001C58C00A88401E (14.99s)
```

danielpaulus:
You're right, @outcast36 — those two are unrelated to your change. Both `TestAccessibilityAudit` and `TestWebInspectorBrowserControl` are failing **only** on device `00008110-001C58C00A88401E`, which is our iOS 26 device — they've been red on that specific device across several unrelated PRs today. It's a known WebInspector/CDP + accessibility-audit problem on iOS 26 that we still owe a fix or a per-device skip; it isn't anything in this PR. The tunnel-free suite and the supervised-wifi paths all passed, so device CI isn't a real blocker here.

Thanks for digging into it 🙏 — I'll follow up with a couple of tiny review notes on the code itself shortly.

danielpaulus:
Thanks again @outcast36 — the security side of this is solid (reuses the vetted `AddProfileSupervised`, no secrets logged). Two things to tighten up and then this is good to merge.

### 1. `isSupervised()` — let's make it shape-agnostic so we can merge without a supervised device
Your move to the nested `cfg["CloudConfiguration"]["IsSupervised"]` is a reasonable inference from how `prepare.go` *writes* the config — but we can't actually confirm the `GetCloudConfiguration` **response** shape right now: the existing `prepare.go`/`erase.go` code fetches that config but never reads a field out of it, there are no fixtures, and **we don't have a supervised device on the CI runners** to check against. (Filed #839 to add one so supervised-only paths like this get real e2e coverage going forward.)

Rather than commit to one shape, could you make the read work either way? That's strictly safer than both the old (top-level only) and new (nested only) versions, and lets us merge this without waiting on hardware:

```go
func isSupervised(conn *Connection) bool {
    cfg, err := conn.GetCloudConfiguration()
    if err != nil {
        return false
    }
    if cc, ok := cfg["CloudConfiguration"].(map[string]interface{}); ok {
        if s, ok := cc["IsSupervised"].(bool); ok {
            return s
        }
    }
    if s, ok := cfg["IsSupervised"].(bool); ok { // fallback: top-level
        return s
    }
    return false
}
```

### 2. A few logging bugs (our `golog` convention)
The new log lines don't match the house rule, and a couple render `!BADKEY` because a value is passed without a key. `golog` is slog-style — every arg after the message must be a `"key", value` pair, and every line should carry `"module", logModule` first plus `"udid", ...` when a device is in scope.

- `golog.Info("PrepareWifi: device supervised: ", supervised)` — `supervised` is a lone bool → renders `... !BADKEY=true`, and it's missing `module`/`udid`. →
  ```go
  golog.Info("device supervised", "module", logModule, "udid", device.Properties.SerialNumber, "supervised", supervised)
  ```
- `golog.Warn("device is supervised, but supervision credentials were not given.", "The Wi-Fi profile must be approved manually...")` — the second string is a lone value → `!BADKEY`. Fold the guidance into the message (or a keyed attr) and add `module`/`udid`.
- `golog.Info("Successfully installed wifi profile")` — add `"module", logModule` and `"udid", ...` (an `"ssid", ...` attr would be a nice touch too).

Everything else looks good. The only red CI check is the unrelated iOS 26 device flake we discussed, so once these two are in, this is merge-ready. 🙏

outcast36:
Apologies for the delay @danielpaulus. Committed fallback method in device is supervised check. Also fixed golog usage.  Only functional difference since opening is the isSupervised checks.

PR #843, opened 2026-08-26
feat(hid): touch, button and keyboard input over CoreDevice HID (builds on #845)

sakhisheikh:
> **Review #845 first.** `ios/display` has been split out into #845 so it can be reviewed on its own. This branch still contains it, because `Session` imports it and the branch has to build — but those 846 lines are the same commit and need no second review. Once #845 merges, this diff shrinks to the HID work on its own (`ios/hid`, `Session`, the CLI) with no rebase needed here.

Adds touch, hardware button and virtual keyboard input over the CoreDevice HID services, so `go-ios` can drive a device without WebDriverAgent.

```
ios hid services
ios hid tap 32767 32767
ios hid drag 45000 40000 20000 40000 --steps 8 --duration 0.1
ios hid button 12 233
ios hid type "hello"
ios hid session --script gestures.txt
```

## The part that is not obvious

Touch reports are gated on a running media stream. Without one, `dtuhidd` publishes the digitizer surfaces as `eventSource: externalAccessory` and `backboardd` drops every event with *"ignoring digitizer event for display &lt;main&gt; from unsupported service"*. The reports are still **accepted without error**, so a caller sees success while nothing happens on screen. Starting a video stream flips the surfaces to authenticated and the same reports arrive as real `UIEventTypeTouches`.

The stream's RTP payload is irrelevant, it only has to exist, so `ios/display` implements stream negotiation and nothing about decoding video: the socket is bound and drained purely so the device keeps sending. The gate is per client: a stream Xcode started does not authenticate our reports.

pymobiledevice3 arrived at the same design independently, which is worth reading if this seems like an odd amount of machinery for a tap: [`hid_service.py`](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/core_device/hid_service.py) documents the gate in its "Authentication gate" comment (`authenticated: NO; builtIn: NO; eventSource: externalAccessory`), and its `touch_session` helper starts a stream, drains the RTP in a background task and notes "we just need a session to exist for the duration of the gestures". It uses the same 1MB receive buffer and the same post-start settle delay, for the same reasons.

`Session` owns this. It opens one connection, starts a stream lazily on the first gesture that needs one, and holds it for the session's lifetime. Negotiating per gesture costs seconds and has been seen to wedge the device's mediastream daemon until reboot. Hardware buttons never start a stream, since that surface is already authenticated.

## Constraints

- **iOS 27 or later.** The services connect from iOS 17, but earlier versions will not start a stream: 26.3 reports no supported media features and fails with CoreDevice error 9021, "Remote control requires iOS 27.0 or later on this device".
- **Kernel tunnel only.** The device sends RTP to a host UDP socket, which a userspace tunnel cannot route. `NewSession` succeeds and the first gesture returns `display.ErrUserspaceTunnelUnsupported`.

## Layout

| package | role |
|---|---|
| `ios/hid` | HID report encoding, the two service wrappers, and `Session` |
| `ios/display` | `mediastreamstart` negotiation, and a UDP sink that discards frames |
| `cmd_device_hid.go` | the `ios hid` subcommands |

The negotiator offer was reverse engineered from `-[AVCMediaStreamNegotiator createOffer]`; the daemon rejects any deviation with an opaque "Invalid Parameter", so `offer_test.go` pins the encoder against a captured Xcode offer byte for byte. `offer.go` documents the blob layout.

Verified on device: gestures, typing and buttons on iOS 27.0, and the 9021 refusal on 26.3.

### DIFF
--- a/cli_device_resolution.go
+++ b/cli_device_resolution.go
@@ -91,6 +91,7 @@ func needsAutomaticTunnelInfo(args docopt.Opts) bool {
 
 	for _, commandName := range []string{
 		"debug",
+		"hid",
 		"devicestate",
 		"instruments",
 		"kill",
--- a/cmd_device_hid.go
+++ b/cmd_device_hid.go
@@ -1,0 +1,362 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/danielpaulus/go-ios/ios/hid"
+	log "github.com/sirupsen/logrus"
+)
+
+// Default pacing for drag gestures. Eight samples over 100ms reads as a flick to
+// the device's gesture recogniser, which is what VoiceOver navigation needs; pass
+// a longer duration for a slow drag.
+const (
+	defaultDragSteps    = 8
+	defaultDragDuration = 100 * time.Millisecond
+)
+
+// runHIDCommand dispatches the `ios hid` subcommands, which inject touch, button
+// and keyboard events through the CoreDevice HID services.
+//
+// Touch input needs a media stream to hold dtuhidd's auth gate open, which
+// hid.Session starts on demand. Every single-gesture command below opens its own
+// session and therefore negotiates its own stream: convenient, but a stream costs
+// about a second to set up, and the device's mediastream daemon has been seen to
+// wedge when many are started in quick succession. Use `ios hid session` to run a
+// batch of gestures through one stream.
+func runHIDCommand(ctx commandContext) {
+	switch {
+	case boolArg(ctx.Args, "services"):
+		runHIDListServices(ctx)
+	case boolArg(ctx.Args, "tap"):
+		runHIDTap(ctx)
+	case boolArg(ctx.Args, "drag"):
+		runHIDDrag(ctx)
+	case boolArg(ctx.Args, "button"):
+		runHIDButton(ctx)
+	case boolArg(ctx.Args, "type"):
+		runHIDType(ctx)
+	case boolArg(ctx.Args, "session"):
+		runHIDSession(ctx)
+	}
+}
+
+func runHIDListServices(ctx commandContext) {
+	withHIDSession(ctx, func(session *hid.Session) error {
+		services, err := session.ListServices()
+		if err != nil {
+			return err
+		}
+		fmt.Println(convertToJSONString(services))
+		return nil
+	})
+}
+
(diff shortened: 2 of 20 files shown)


sakhisheikh:
Opened #844 to discuss whether you want this at all, and whether you would rather review it as four smaller stacked PRs. Either way, here is a reading order so this does not have to be read as 3.2k lines.

Most of the diff is mechanical: report builders pinned by golden byte tests against pymobiledevice3's reference output, and a CLI. The genuinely novel parts are about 300 lines:

1. `ios/hid/hid.go`, package doc. Why touch needs a media stream at all: dtuhidd publishes the digitizer surfaces as `externalAccessory` and backboardd discards every report aimed at one, without an error, so a caller sees success while nothing happens. Nothing else in the PR makes sense without this.
2. `ios/display/offer.go`, the comment block at the top. The `negotiatorOffer` wire format, reverse engineered from `-[AVCMediaStreamNegotiator createOffer]`. The daemon rejects any deviation with an opaque "Invalid Parameter", which is why `offer_test.go` pins the encoder byte for byte against a captured Xcode offer.
3. `ios/hid/session.go`, the `Session` doc and `ensureStream`. One stream held for the session's lifetime rather than per gesture, started lazily, plus the teardown order: a held contact is lifted before the stream goes, or the device keeps believing a finger is on the screen.

Everything else follows from those three.

Two things I would rather flag than have you find:

`surfaceSettleDelay` is a 300ms sleep after the stream starts. I tried to replace it with a real check and could not: I verified on device that surface 257 never reports `Authenticated`, even with a stream up and touches landing, so the host has no signal to poll. The transition backboardd logs is on its own `BKIOHIDService` objects, not in anything `ListConnectedServices` returns.

`SurfaceMainTouchscreen = 257` is a constant rather than resolved from `ListConnectedServices`. Confirmed correct on iOS 27 (`dtuhidd` logs `mainTouchscreen(0x101)`), but it is an assumption.

Verified on device: gestures, typing and buttons on iOS 27.0; the 9021 refusal on 26.3; and 1860 touch reports delivered with zero drops at sustained 26-61 reports per second.

sakhisheikh:
Parking this so there is only one thing to look at. #845 is the first half, `ios/display` on its own, and it is ready for review.

I will reopen this with the HID half once #845 lands, at which point the diff here is only `ios/hid`, `Session` and the CLI, roughly 2.4k lines instead of 3.2k. Nothing is lost by closing it: the branch stays, and the reading order and context above still apply.

Context for both is in #844.

PR #845, opened 2026-08-28
feat(display): CoreDevice media stream negotiation

sakhisheikh:
Adds `ios/display`, which drives `com.apple.coredevice.displayservice`. the service behind Xcode's device screen view. It negotiates and starts/stops an RTP video stream of a display. 

**Supported OS versions:** iOS27+

## Implementation
- We ported the implementation mostly from pymobiledevice3 which documents it in [`media_stream_offer.py`](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/core_device/media_stream_offer.py).

- `Receiver` binds a UDP socket and drains it without decoding. The drain loop is not optional. Without a reader the socket buffer fills, the device sees packet loss and throttles its encoder, which shows up as stream stalls. pymobiledevice3 does the same thing for the same reason, with the same 1MB receive buffer: see the `touch_session` helper in [`hid_service.py`](https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/core_device/hid_service.py).

- Kernel tunnel only. The device sends RTP back to the host, and the userspace stack registers TCP only.

**Testing:** Verified on an iPhone running iOS 27.0: streams start, stop cleanly, and hold the HID auth gate open for the duration.

### DIFF
--- a/ios/display/display.go
+++ b/ios/display/display.go
@@ -1,0 +1,190 @@
+// Package display starts RTP video streams over
+// com.apple.coredevice.displayservice. A stream also authenticates HID input.
+package display
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"sync"
+
+	"github.com/danielpaulus/go-ios/ios"
+	"github.com/danielpaulus/go-ios/ios/coredevice"
+	"github.com/danielpaulus/go-ios/ios/xpc"
+	"github.com/google/uuid"
+)
+
+const logModule = "go-ios/display"
+
+const (
+	serviceName = "com.apple.coredevice.displayservice"
+
+	featureStartMediaStream = "com.apple.coredevice.feature.startmediastream"
+	featureStopMediaStream  = "com.apple.coredevice.feature.stopmediastream"
+
+	actionMediaStreamStart = "com.apple.coredevice.action.mediastreamstart"
+	actionMediaStreamStop  = "com.apple.coredevice.action.mediastreamstop"
+
+	// Host capabilities and link type. The values Xcode sends over a tunnel, with
+	// no documented meaning to derive them from.
+	clientSupportedFeatures = uint64(140)
+	accessNetworkType       = int64(1)
+	transportProtocolType   = int64(2)
+
+	// The deadline the device applies to its own side. Keep it under the caller's:
+	// the device giving up first answers with an error and keeps the connection.
+	negotiationTimeoutSeconds = uint64(10)
+
+	DefaultDisplayID = 1
+)
+
+// Service is a connection to the device's display service. One request runs at a
+// time: nothing in the XPC framing pairs a reply with its request (see invoke).
+type Service struct {
+	conn     xpcConn
+	deviceID string
+	mutex    sync.Mutex
+}
+
+// xpcConn is the subset of *xpc.Connection used here, so the request/reply logic
+// can be exercised with a fake.
+type xpcConn interface {
+	Send(data map[string]interface{}, flags ...uint32) error
+	ReceiveOnServerClientStream() (map[string]interface{}, error)
+	Close() error
+}
+
+// New connects to the display service. The Developer Disk Image must be
+// mounted, otherwise the service is not published and connecting fails.
+func New(device ios.DeviceEntry) (*Service, error) {
+	conn, err := ios.ConnectToXpcServiceTunnelIface(device, serviceName)
+	if err != nil {
+		return nil, fmt.Errorf("display.New: failed to connect to %s: %w", serviceName, err)
+	}
+	return &Service{conn: conn, deviceID: uuid.New().String()}, nil
+}
+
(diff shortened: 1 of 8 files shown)


sakhisheikh:
How a media stream is started, touch events are part of subsequent PR. it's shown here as how it would work along with the media stream service:

```mermaid
sequenceDiagram
    participant H as Host (go-ios)
    participant D as Device (iOS 27)

    Note over H: bind a UDP socket for the video
    Note over H: read and discard it, we need the<br/>stream to exist, not the pictures

    H->>D: XPC: start a video stream (+ offer)
    D-->>H: accepted, here is the session id
    D->>H: video arrives (RTP/H.264, continuous)

    Note over H: wait 300ms so the device is still<br/>switching over, taps sent now vanish
    Note over D: touch input from us now counts as<br/>the real touchscreen, not an accessory

    H->>D: touch reports → real taps on screen

    H->>D: XPC: stop the stream
```

ios/display/display.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, fmt.Errorf("display.New: failed to connect to %s: %w", serviceName, err)
+	}
+	return &Service{conn: conn, deviceID: uuid.New().String()}, nil
+}
+
+// Close closes the connection to the display service. Stop any running stream
+// first, otherwise the device keeps sending RTP until it times out.
+func (s *Service) Close() error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
```
buxomant: This is the same mutex held by `invoke` (for its entire duration), so if `invoke` hangs we cannot interrupt it using `Close`? I'm not sure how this `Close` is expected to be used, or if the mutex is all that useful as written.

ios/display/display.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := s.conn.Send(request, xpc.HeartbeatRequestFlag); err != nil {
+		_ = s.conn.Close()
+		return nil, fmt.Errorf("display: failed to send %s: %w", feature, err)
+	}
+
+	response, err := s.conn.ReceiveOnServerClientStream()
+	if err != nil {
+		_ = s.conn.Close()
+		if ctx.Err() != nil {
+			return nil, fmt.Errorf("display: %s did not complete: %w (the device's mediastream daemon may be wedged; rebooting the device clears it)", feature, err)
```
buxomant:
```suggestion
			return nil, fmt.Errorf("display: %s did not complete: %w (the device's mediastream daemon may be wedged; rebooting the device clears it)", feature, ctxErr)
```

ios/display/display.go
```diff
@@ -0,0 +1,191 @@
...
+		"senderIP":     req.SenderIP,
+		"timeout":      negotiationTimeoutSeconds,
+		"type":         "video",
+	}
+
+	output, err := s.invoke(ctx, featureStartMediaStream, actionMediaStreamStart, input)
+	if err != nil {
+		// The id is returned even on failure: the device may have started the
+		// stream regardless (or be about to), and stopping it needs this id.
+		return StreamAnswer{ClientSessionID: clientSessionID}, err
```
buxomant: Depending on why the `StartVideoStream` call failed (e.g. in case of timeout), the caller might need to try to create a new `Service` to call `StopMediaStream`. Otherwise we'll try to use the same (broken) service to  Maybe worth a code comment? (this was flagged by Claude and may be overkill imo)

sakhisheikh: i think that's a good point. Updating the comment.

ios/display/display.go
sakhisheikh: yes that makes sense

ios/display/display.go
sakhisheikh: it was added before i refactored the tear down with `context`. We were using in memory state to determine if the session was abandoned and mutex was used for that. 

>  if the mutex is all that useful as written.

Mutex is useful in `invoke` to keep one request at a time that's how XPC is designed. We don't have unique identifier for match a reply with its request.

ios/display/display.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil, fmt.Errorf("display.New: failed to connect to %s: %w", serviceName, err)
+	}
+	return &Service{conn: conn, deviceID: uuid.New().String()}, nil
+}
+
+// Close closes the connection to the display service. Stop any running stream
+// first, otherwise the device keeps sending RTP until it times out.
+//
+// Deliberately not behind the mutex: invoke holds that for the whole exchange,
+// so taking it here would stop Close from interrupting a call that is hanging.
```
buxomant: Minor, but I don't think we should generally explain why we're _not_ doing something. There are many things that we don't want to do (because they would be a bad idea), it does not add much value to document all of them.

sakhisheikh: yes I found out this comment later and i also didn't like it.

buxomant (approved):

PR #848, opened 2026-09-01
fix(imagemounter): tolerate a device reporting an image already mounted

aluedeke:
## Problem

`MountImage()` checks `ListImages()` first and skips the mount if anything is already reported mounted — but on iOS 17+, a personalized DDI auto-remounts within about a second of boot (Developer Mode persists it across reboots, unlike the older mount-once-per-session model). That's fast enough to lose the race against `ListImages()`: it can observe nothing mounted, and the device still refuses the immediately-following mount attempt as a duplicate.

Reproduced directly against a real device (iOS 26.6), including right after a full reboot:

```
ListImages() -> none
MountImage() -> ImageMountFailed: ... "A disk image of type Personalized/DeveloperDiskImage is already mounted at /System/Developer."
```

An explicit `UnmountImage()` also fails in this state (`Failed to unload launchd jobs`), confirming something is genuinely mounted despite `ListImages()` saying otherwise — this isn't a false failure to swallow blindly, it's a real detection race between two different device queries.

## Fix

Treat that specific device response the same way the existing `ListImages()` pre-check already does: log a warning and return `nil` instead of failing the whole mount. Scoped narrowly via a new `alreadyMounted(err)` helper that only matches `ImageMountFailed` errors whose detail contains "already mounted at" — any other mount failure (developer mode disabled, locked device, wrong signature, ...) still fails normally.

## Testing

- `TestAlreadyMounted` (new): matches the real device-captured "already mounted" response, and confirms other error shapes (developer-mode-disabled, device-locked, nil, generic errors) are not matched.
- `go build ./...` and `go test ./ios/imagemounter/...` pass, aside from `TestUsesProxy`/`TestWorksWithoutProxy`, which are pre-existing failures unrelated to this change — they call the mounter's `MountImage` directly (bypassing the package-level `MountImage()` wrapper this PR touches) and are dependent on whatever real device happens to be attached to the machine running them.

### DIFF
--- a/ios/imagemounter/imagemounter.go
+++ b/ios/imagemounter/imagemounter.go
@@ -242,7 +242,28 @@ func MountImage(device ios.DeviceEntry, path string) error {
 		golog.Warn("there is already a developer image mounted, reboot the device if you want to remove it. aborting.", "module", logModule, "udid", device.Properties.SerialNumber, "imagePath", path)
 		return nil
 	}
-	return conn.MountImage(path)
+
+	err = conn.MountImage(path)
+	if alreadyMounted(err) {
+		// A personalized DDI can auto-remount within about a second of boot (part of
+		// Developer Mode persisting across reboots) — faster than ListImages() above
+		// can catch, so the device can still refuse this as a duplicate right after
+		// it just reported nothing mounted. Treat it the same as the check above.
+		golog.Warn("device reports an image already mounted; treating as already in the desired state.", "module", logModule, "udid", device.Properties.SerialNumber, "imagePath", path)
+		return nil
+	}
+	return err
+}
+
+// alreadyMounted reports whether err is the device refusing MountImage because an image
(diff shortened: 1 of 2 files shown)


ios/imagemounter/imagemounter.go
```diff
@@ -242,7 +242,28 @@ func MountImage(device ios.DeviceEntry, path string) error {
 		golog.Warn("there is already a developer image mounted, reboot the device if you want to remove it. aborting.", "module", logModule, "udid", device.Properties.SerialNumber, "imagePath", path)
 		return nil
 	}
-	return conn.MountImage(path)
+
+	err = conn.MountImage(path)
+	if alreadyMounted(err) {
```
gmegidish, line 247: I would suggest renaming this to isAlreadyMountedErr

gmegidish (approved):

PR #849, opened 2026-09-03
feat(hid): touch input over CoreDevice HID

sakhisheikh:
Adds `ios/hid`, which injects touch events into an iOS device over CoreDevice.

**A media stream has to be running or nothing happens.** Without one the device
publishes the touch surfaces as an external accessory and throws every report
away, while the send still succeeds. The caller sees no error and nothing moves
on screen. The stream is per client, so an Xcode someone else has open does not
count.

Starting that stream is deliberately **not** this package's job. An earlier
revision owned it here, and the review asked why a HID package owns a video
stream; the answer was only that the ordering check had to live somewhere it
could not be skipped. It now lives in the caller, which has its own single
entry point for touches, and `ios/hid` connects and sends reports.

Two things a caller has to get right, both documented on the exported API:

- Start a media stream before sending anything, and keep it up for the session.
  `ios/display` does this.
- A nil error means the report was written, not that anything happened. The
  device never acknowledges one.

`Session` tracks whether a contact is down so `Close` can lift it. Without that
an interrupted session leaves the device believing a finger is on the screen,
which nothing recovers from.

## Requirements

**iOS 27 and a kernel tunnel.** Below 27 the service is not advertised in RSD at
all: three devices checked, builds `24A5355q` and `24A5370h` list 76 services
with no HID entry, `24A5390f` lists 85 including
`com.apple.coredevice.hid.universalhidservice`. An iOS 18 device refuses with
CoreDevice error 9021, checked on an iPhone16,1 running 18.7.3.

The Developer Disk Image matters: `dtuhidd`, which serves that endpoint, ships
inside the image rather than in the OS, so an older DDI mounts fine and simply
does not provide it.

Verified end to end on an iPhone14,6 running iOS 27.0 build `24A5390f`, and
through a full live stack on staging hardware.

## Also in here

`ios/display` loses `StreamAnswer`, whose only remaining field was the session
id, so `StartVideoStream` returns the id directly. `Receiver.Drain` is removed;
its one caller now uses `io.Copy(io.Discard, receiver)`, which says at the call
site that the frames are thrown away.

A Go port of the HID work in pymobiledevice3:
https://github.com/doronz88/pymobiledevice3/blob/master/pymobiledevice3/remote/core_device/hid_service.py


sakhisheikh:
How a session works. The caller owns the media stream; `ios/hid` connects and
sends reports.

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as display.Service
    participant S as hid.Session
    participant P as Device (iOS 27)

    C->>D: OpenReceiver, then StartVideoStream
    D->>P: start the stream
    Note over P: touch from us now counts as<br/>the real touchscreen
    Note over C: about a second, once per session

    C->>S: NewSession
    S->>P: connect to the HID service

    C->>S: TouchDown, TouchMove, TouchUp
    S->>P: one report per call, applied as it arrives

    C->>S: Close
    S->>P: lift a contact still down
    C->>D: StopMediaStream
```

The order matters in both directions. The stream goes up before the first
report, because reports sent before the surfaces switch over are discarded with
no error. On the way down the contact is lifted before anything closes, or the
device keeps believing a finger is on the screen.

sakhisheikh:
### Where to look

**1. Why the caller has to run a video stream.** [hid.go](https://github.com/sakhisheikh/go-ios/blob/8616184d1b54b0b66b0d950c2b9f0da908330faf/ios/hid/hid.go#L1)

Touch reports are accepted and thrown away unless a stream is running, and the
send still returns nil. Everything else in the package follows from that.

**2. What a Session owns.** [session.go#L32](https://github.com/sakhisheikh/go-ios/blob/8616184d1b54b0b66b0d950c2b9f0da908330faf/ios/hid/session.go#L32)

A connection and one held contact. No stream, no receiver, no goroutine: an
earlier revision had all of those and the review asked why. Worth asking as you
read: who lifts the finger if the client disappears.

**3. The ordering in Close.** [session.go#L116](https://github.com/sakhisheikh/go-ios/blob/8616184d1b54b0b66b0d950c2b9f0da908330faf/ios/hid/session.go#L116)

The contact is released before anything closes. Reverse it and the connection
is gone before the release can be sent.

**4. Why the context is ignored in TouchUp.** [session.go#L76](https://github.com/sakhisheikh/go-ios/blob/8616184d1b54b0b66b0d950c2b9f0da908330faf/ios/hid/session.go#L76)

Honouring cancellation there would skip the release exactly when the caller is
tearing down, which is when a stuck finger matters most.

### Two things that are easy to get wrong

`TouchDown` and `TouchMove` send the identical report. The device tracks where
the contact is, not what changed, so a move is a contact at a new point.

A nil error means the report was written, not that anything happened. Nothing
acknowledges it.

### The rest

`payload.go` builds the request dictionary, `report.go` packs bytes to the
layout in its comments, and the tests are the evidence for everything above.

ios/hid/session.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+	s.displayService = svc
+
+	// Start draining before the stream is negotiated: the device begins sending
+	// as soon as it answers, and an unread socket makes it throttle its encoder.
+	s.streamLost.Store(false)
+	s.drainDone = make(chan struct{})
+	go func(done chan struct{}) {
+		defer close(done)
+		if err := receiver.Drain(); err != nil {
```
dmissmann: Drain is doing essentially what usually is done with `_, err := io.Copy(io.Discard, s.receiver)` (and it makes it clear that this is thrown away)

And does this mean that with this implementation we lock ourselves out of consuming the video stream?

ios/hid/session.go
```diff
@@ -0,0 +1,446 @@
...
+		SenderIP:     s.device.Address,
+		DisplayID:    display.DefaultDisplayID,
+	})
+	s.streamAnswer = answer
+	if err != nil {
+		s.teardownStream()
+		return fmt.Errorf("failed to start the media stream touch input requires: %w", err)
+	}
+
+	golog.Info("media stream started, touch surfaces authenticated", "module", logModule,
```
dmissmann: what does this authentication look like? Is there anything at all used from this connection here?

ios/hid/hid.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if err := c.conn.Send(buildSendReportPayload(serviceID, report), xpc.HeartbeatRequestFlag); err != nil {
+		return fmt.Errorf("SendReport: failed to send report to surface %d: %w", serviceID, err)
+	}
+	return nil
+}
+
+// SendTouchscreen posts one touchscreen report at (x, y). Every TouchContact
+// means "in contact here", so a drag is a run of them ending in TouchRelease.
+func (c *UniversalConnection) SendTouchscreen(state TouchState, x, y uint16, serviceID uint64) error {
```
dmissmann: `serviceID` tells what's the target or what is it for? And if that's the case,, isn't that part of the method name already? (and the name also sounds a bit weird. we rather send a touch event that a touchscreen)
```suggestion
func (c *UniversalConnection) SendTouchscreen(state TouchState, p Point, serviceID uint64) error {
```

ios/hid/hid.go
```diff
@@ -0,0 +1,71 @@
...
+// and the Developer Disk Image must be mounted so dtuhidd is running.
+func NewUniversal(device ios.DeviceEntry) (*UniversalConnection, error) {
+	conn, err := ios.ConnectToXpcServiceTunnelIface(device, universalServiceName)
+	if err != nil {
+		return nil, fmt.Errorf("NewUniversal: %w", err)
+	}
+	return &UniversalConnection{conn: conn}, nil
+}
+
+func (c *UniversalConnection) ListConnectedServices() (map[string]interface{}, error) {
```
dmissmann: What's this for? It's only used in a test, and there we check that there's an error

ios/hid/hid.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	res, err := c.conn.ReceiveOnServerClientStream()
+	if err != nil {
+		return nil, fmt.Errorf("ListConnectedServices: failed to read response: %w", err)
+	}
+	return res, nil
+}
+
+// The device never returns a response, so a nil error means it was sent, not that anything moved.
+func (c *UniversalConnection) SendReport(serviceID uint64, report []byte) error {
```
dmissmann:
```suggestion
func (c *UniversalConnection) sendReport(serviceID uint64, report []byte) error {
```

and where does the name `report` come from? is that also used by Apple?

ios/hid/report.go
```diff
@@ -0,0 +1,66 @@
...
+// BuildTouchscreenReport carries a contact state and position.
+// Layout: [0]=report ID, [1:3]=constants 0x01 0x05, [3]=state, [4:6]=X uint16 LE,
+// [6:8]=Y uint16 LE, [8:40] reserved, [40:44]=constant 0x02 0x00 0x00 0x00,
+// [44:50]=timestamp, [50:58] reserved.
+//
+// A contact at x=0x1234 y=0x5678 with timestamp 0xa1b2c3d4:
+//
+//	09 0105 c2 3412 7856 00..00 02000000 d4c3b2a10000 00..00
+//	ID const st x    y    res    const    timestamp    res
+func BuildTouchscreenReport(state TouchState, x, y uint16, ts uint64) []byte {
```
dmissmann:
```suggestion
func buildTouchscreenReport(state TouchState, x, y uint16, ts uint64) []byte {
```

ios/hid/report.go
```diff
@@ -0,0 +1,66 @@
...
+	TouchRelease TouchState = 0x02
+)
+
+// processStart anchors Timestamp to a monotonic origin. time.Since reads Go's
+// monotonic clock, so the sequence is unaffected by wall-clock adjustments.
+var processStart = time.Now()
+
+// Timestamp returns the value to stamp a report with. Only monotonicity and the
+// deltas between reports are read from it, not absolute time.
+func Timestamp() uint64 {
```
dmissmann:
```suggestion
func timestamp() uint64 {
```

ios/hid/report.go
```diff
@@ -0,0 +1,66 @@
...
+	"time"
+)
+
+const (
+	reportIDTouchscreen = 0x09
+)
+
+// Report lengths. The device rejects reports of any other size for these surfaces.
+const (
+	TouchscreenReportLen = 58
```
dmissmann:
```suggestion
	touchscreenReportLen = 58
```

ios/hid/session.go
```diff
@@ -0,0 +1,447 @@
...
+
+	mutex sync.Mutex
+	hid   hidConn
+
+	// Stream state, populated by ensureStream and guarded by mutex, except
+	// streamLost which the drain goroutine writes without it.
+	displayService streamStopper
+	receiver       *display.Receiver
+	streamAnswer   display.StreamAnswer
+	drainDone      chan struct{}
```
dmissmann, line 70: is that used properly? I see that we expect a message somewhere, but where's the write?

sakhisheikh: it is in the goroutine right below where it is passed and on error it is closed with defer.

> 	go func(done chan struct{}) {
> 		defer close(done)
> 		if err := receiver.Drain(); err != nil {
> 			golog.Warn("the media stream that gates touch input stopped",
> 				"module", logModule, "error", err)
> 			s.streamLost.Store(true)
> 		}
> 	}(s.drainDone)

ios/hid/session.go
sakhisheikh: I can grab the log line, it receives the error response payload which doesn't allow the touch to go through

ios/hid/session.go
sakhisheikh: Yes but we have both options here to actually also read the frame if we would like to.

ios/hid/session.go
sakhisheikh: authenticates mean if the request is allowed, it gets errored instead, Maybe i can reword it.

ios/hid/hid.go
sakhisheikh: Yes it is, I validated  claude claims and apple corehid package has the convention: https://developer.apple.com/documentation/corehid/hiddeviceclient/dispatchsetreportrequest(type:id:data:timeout:)

ios/hid/hid.go
sakhisheikh: deleted.

ios/hid/session.go
sakhisheikh: I mean media stream gates the touches to device hub. what wording would have made sense?

ios/hid/session.go
sakhisheikh: But i removed drain and added more explicit draining  with `_, err := io.Copy(io.Discard, s.receiver)`.

We have `Read` exported method that we could use to consume the rtp packets.

ios/hid/session.go
sakhisheikh: reworded it.

sakhisheikh:
@dmissmann I have removed unnecessary exports and removed unused and redundant methods. 

- Now a consumer will need to call only these three method to perform a touch or a drag. a touch is only down and then up while a drag or a gesture is down xN and then up.
```
case EventDown: return s.hid.TouchDown(ctx, p)
case EventMove: return s.hid.TouchDown(ctx, p)
case EventUp:   return s.hid.TouchUp(ctx, p)
```

- also i decoupled display stream code from hid completely, It's up to the caller how to use it in combination. 

- lost stream will return an error instead of keep trying to reconnect always as i couldn;t reproduce it that if in mid session we lose the stream somehow>

ios/hid/session.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	session.hid = conn
+	return session, nil
+}
+
+// TouchDown reports a contact at point and leaves it there. Call it again at a
+// new point to move: the device tracks where the contact is, not what changed,
+// so there is no separate move. TouchUp lifts it.
+func (s *Session) TouchDown(ctx context.Context, point Point) error {
+	return s.sendContact(ctx, point, "TouchDown")
```
dmissmann: you can move `sendContact` in here now, there are no other callers

dmissmann (approved):

ios/hid/session.go
sakhisheikh: inlined it.

PR #855, opened 2026-09-22
Add LiquidGlass to skipAllSetup

thefrank60:
iOS 26 introduced a "Liquid Glass" pane in Setup Assistant -- the design-language intro with a clear-to-tinted slider. It has a SkipSetupItems key, "LiquidGlass", but that key is missing from skipAllSetup, so `ios prepare --skip-all` does not skip it.

The effect is easy to misread: Prepare() still returns successfully, so prepare looks like it worked, but the device stays in Setup Assistant parked on that one pane with every later pane already skipped. Setup never completes without a physical tap on Continue.

Verified on an iPhone19,3 running iOS 27.0 (build 24A427): erase, activate, then `prepare --skip-all`. Without the key the device sits on the Liquid Glass pane indefinitely. With it, SetupDone and SetupFinishedAllSteps both become true roughly 60 seconds after prepare returns, with no interaction. Reproduced on three devices.


davidmathis:
LGTM - thanks @thefrank60 

cc: @danielpaulus

aircraft-cerier:
Thanks for adding this! ✅

thefrank60:
@buxomant can you please take a look, thank you.

buxomant (approved):

